PR #37, opened 2014-11-01
Default dialog position changed to centered over main app window; Internationalization and German translation

jimmytaker:
Changes as discussed.


vslavik:
I uploaded the German translation to https://crowdin.com/project/winsparkle where I'll manage WinSparkle translations; I opted to do the code slightly differently.

Issue #39, opened 2015-01-02
Optionally launch web browser


joshkel:
Our updates are hosted on a password-protected web site, and rather than make users enter their username and password into the WinSparkle UI, I'd like to simply launch a web browser to the download page if a new version is detected.

It looks like neither WinSparkle nor Sparkle currently supports this?

Any thoughts on adding such a feature? E.g., if no `<enclosure>` is present and a `<link>` is present, then launch a browser to that link if the user clicks update. I can work on implementing it if you're open to the idea.

vslavik:
> E.g., if no <enclosure> is present and a <link> is present, then launch a browser to that link if the user clicks update

Sounds sensible to me.

vslavik:
> I can work on implementing it if you're open to the idea.

@joshkel Just to be clear: yes, I think it makes sense to add it, and a pull request would very much be appreciated. (I just realized my in-a-hurry reply above was a bit terse.)

joshkel:
@vslavik I understood, and I just submitted a pull request. Thanks!

PR #40, opened 2015-01-09
Add support for visiting web site instead of download.

joshkel:
If an RSS entry has a <link> attribute with a valid version number and has no
<enclosure>, then WinSparkle will launch a web browser to that link rather
than downloading and installing the update itself. This will make it easier to
use WinSparkle with password-protected downloads.

Closes vslavik/winsparkle#39.

### DIFF
--- a/src/appcast.cpp
+++ b/src/appcast.cpp
@@ -58,6 +58,8 @@ namespace
 #define ATTR_VERSION    NS_SPARKLE_NAME("version")
 #define ATTR_SHORTVERSION NS_SPARKLE_NAME("shortVersionString")
 #define ATTR_OS         NS_SPARKLE_NAME("os")
+#define NODE_VERSION      ATTR_VERSION        // These can be nodes or
+#define NODE_SHORTVERSION ATTR_SHORTVERSION   // attributes.
 #define OS_MARKER       "windows"
 
 // context data for the parser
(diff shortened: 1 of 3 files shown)


vslavik:
I'm done with code review nitpicking (this is overall *very* good, so really: all that above is just nitpicking).

There's one bigger issue, though, that I didn't even consider before until I tripped over `ParseVersionAttr`:  the fields that provide sparkle version information are defined for the *enclosure*. Your code interprets them on `<link>` elements too and I'm a bit uneasy about intentionally introducing spec (even if loose) deviations like that into WinSparkle.

So I did a bit of research and good news — turns out that Sparkle already supports this, via a `<sparkle:version>` element under the item. See sparkle-project/Sparkle#249  (and [the referenced mention by Andy](https://github.com/sparkle-project/Sparkle/issues/227#issuecomment-13192723)), it looks like this:
```xml
<item>
  <title>Version 1.2.4</title>
  <sparkle:releaseNotesLink>http://foo.bar/storage/downloads/macosx/release_notes_test.html</sparkle:releaseNotesLink>
  <pubDate>Mon, 28 Jan 2013 14:30:00 +0500</pubDate>
  <sparkle:version>1.2.4</sparkle:version>
  <link>http://foo.bar/download_link.html</link>
</item>
```

Could you update your patch to use the `<sparkle:version>` element (and `<sparkle:shortVersionString>` too, also supported by Sparkle) under `<item>` instead? TIA!

joshkel:
Thanks for the feedback. I tried to search if Sparkle already did anything
like this and failed to find it, so thank you very much for the info.

I'll make this change and update my pull request.

joshkel:
Updated. Let me know if there are any other changes you'd like to have made. Thanks.

vslavik:
Wonderful, thanks a lot!

> I wondered about trying to consolidate the node logic in OnStartElement, OnEndElement, and OnText but wasn't sure that I should do that as part of this commit.

I can see why, this part of the code is getting increasingly messy. But it's probably not worth the effort, it would be better to rewrite it to use TinyXML or something (or even MSXML) — that would provide friendlier API than Expat for this use *and* be smaller in binary size.

Issue #19, opened 2014-05-27
Support for minimumSystemVersion


vslavik:
I can't seem to get miniumSystemVersion working with winsparkle. I'm trying to only allow updates for users on win7 or above.
```
<sparkle:minimumSystemVersion>6.1</sparkle:minimumSystemVersion>
```
Works for Mac's Sparkle, but not for WinSparkle.

vslavik:
Fixed in d9090367487af9f0186e68cc4087b29ef11618cb.

PR #43, opened 2015-01-17
Add support for minimumSystemVersion

awinslow:
This addresses issue #19  by adding support for the <sparkle:minimumSystemVersion> tag in an appcast.  This allows you to be able to specify an item to only be installable on at least a certain version of windows (and above).  If the tag is not specified on an item it will remain available for all versions as is the case now.  The format for the tag is MajorRelease.MiniorRelease.ServicePackLevel  with the service pack level being optional.  For example, to make an item only available on windows 7 and above add `<sparkle:minimumSystemVersion>6.1</sparkle:minimumSystemVersion>` or for XP SP3 and above, you would use `<sparkle:minimumSystemVersion>5.1.3</sparkle:minimumSystemVersion>`

### DIFF
--- a/src/appcast.cpp
+++ b/src/appcast.cpp
@@ -29,6 +29,7 @@
 #include <expat.h>
 #include <vector>
 #include <algorithm>
+#include <Windows.h>
 
 namespace winsparkle
 {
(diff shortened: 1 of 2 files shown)


vslavik:
Great! Thanks a lot — and thanks for taking the time to make the patch this good.

PR #49, opened 2015-01-23
Prevent downloader from running twice.

kenjiuno:
```
- You can invoke downloader twice, if you click
  [Check for updates manually] and then,
  [Install update] at UpdateDialog.
  Repeat it once more, before downloader completes.
- App hangs due to null pointer access of m_downloader
  at UpdateDialog::StateUpdateDownloaded.
```


vslavik:
Thanks!

PR #50, opened 2015-02-11
Add support for 32/64bit specific items

TingPing:


vslavik:
This seems useful, thanks! Sparkle doesn't deal with bitness at all, so this way of tagging it is as good as any other. 

My only comment in addition to the code-review ones above is that it would be nice to have some human-readable documentation for these extensions, so if you wrote it down in the wiki, I wouldn't mind.

TingPing:
>This comment is worse than the previous one: it is less specific.

Well the docs for ```is_valid_windows_item()``` are where the specifics went. But I can duplicate that text there.

I planned on mentioning it in the wiki after it merged, but I can do that now.

TingPing:
Added a line about it here: https://github.com/vslavik/winsparkle/wiki/Appcast-Feeds

vslavik:
Perfect, thanks a lot!

vslavik:
I just realized something: wouldn't it be beneficial to make the 32/64 distinction based on the OS capability rather than the current binary? I realize that making the decision based on the binary performing the check, as this PR does, may be useful e.g. for people who purposefully install 32bit version instead of 64bit even on 64bit Windows (is there a reason for this?).

But think about another scenario: previously 32bit-only application now has a 64bit version too and would like the updates to migrate over to x64. Shouldn't that be possible to do too?

I think it should, and your 9a1f4d1 makes it trivial to add, I just don't know how should the tag be called: `windows-64` is taken and `windows-wow64` (the only other thing I can think of) would be a misnomer. Would you have any idea how to name these 3 cases (32bit, 64bit client binary, 64bit OS regardless of client bitness) by a chance?

TingPing:
>for people who purposefully install 32bit version instead of 64bit even on 64bit Windows (is there a reason for this?).

There are valid reasons, such as the program uses a scripting language and users want to only maintain 32bit python installs.

>But think about another scenario: previously 32bit-only application now has a 64bit version too and would like the updates to migrate over to x64. Shouldn't that be possible to do too?

It is possible for updates to link to a website not a download right? That would be a good situation to let the user decide.

At the end of the day I think we should never overrule the decision on what bitness their application is.

vslavik:
> At the end of the day I think we should never overrule the decision on what bitness their application is.

I don't know about that. For example, most users of [Poedit](https://github.com/vslavik/poedit) absolutely don't give a damn (or even understand the distinction) and I very much would want to update them to x64 builds when possible; it's not something users should be required to *decide*. In fact I'd say it's the case when the users have external, independently installed dependencies (as is yours) that is the edge case… 

Well, never mind then.

TingPing:
>I'd say it's the case when the users have external, independently installed dependencies (as is yours) that is the edge case…

Perhaps, as long as it is another option it wont bother me.

TingPing:
@vslavik Question, should it be valid to have two enclosures in a single item?

```xml
<item>
    <enclosure sparkle:os="windows-x86" />
    <enclosure sparkle:os="windows-x64" />
</item>
```

vslavik:
> @vslavik Question, should it be valid to have two enclosures in a single item?

I would think so, yes.

PR #58, opened 2015-05-18
mingw compilation fixes

drizt:
Hello.

I am trying to build WinSparkle with MinGW. I'm doing this on Fedora 20 (with MinGW). Also I use CMake file (I'm writing them). At this moment i built libWinSparkle.a. So this pull request contains some code fixes for this (without any CMake files).


vslavik:
Thanks!

PR #60, opened 2015-05-18
fixed warning

drizt:
DWORD_PTR is typdef for integer not pointer.

From MSDN [DWORD_PTR](https://msdn.microsoft.com/en-us/library/windows/desktop/aa383751%28v=vs.85%29.aspx#DWORD_PTR) and [ULONG_PTR](https://msdn.microsoft.com/en-us/library/windows/desktop/aa383751%28v=vs.85%29.aspx#ULONG_PTR). MinGW compiler produces warning when NULL is used for DWORD_PTR.

### DIFF
--- a/src/download.cpp
+++ b/src/download.cpp
@@ -152,7 +152,7 @@ void DownloadFile(const std::string& url, IDownloadSink *sink, int flags)
                           NULL, // lpszHeaders
                           -1,   // dwHeadersLength
                           dwFlags,
-                          NULL  // dwContext
+                          0     // dwContext
                       );
     if ( !conn )
         throw Win32Exception();


vslavik:
*Semantically* it is a pointer, not an integer, and `NULL` is more appropriate.

What *is* the gcc warning anyway?

drizt:
[ 96%] Building CXX object src/CMakeFiles/WinSparkle.dir/download.cpp.obj
    I:\winsparkle\src\download.cpp: In function 'void winsparkle::DownloadFile(const string&, winsparkle::IDownloadSink*, int)':
    I:\winsparkle\src\download.cpp:156:23: warning: passing NULL to non-pointer argument 6 of 'void* InternetOpenUrlA(HINTERNET, LPCSTR, LPCSTR, DWORD, DWORD, DWORD_PTR)' [-Wconversion-null]

PR #61, opened 2015-05-19
Cmake for MinGW

drizt:
It tested with Windows MinGW Qt 5.3.1 and Fedora MinGW.
For MSVC 2010 I can't build. Probably support for MSVC can be added. But I haven't such wishes atm.

### DIFF
--- a/3rdparty/expat/ConfigureChecks.cmake
+++ b/3rdparty/expat/ConfigureChecks.cmake
@@ -40,5 +40,5 @@ else(HAVE_SYS_TYPES_H)
     set(SIZE_T "unsigned")
 endif(HAVE_SYS_TYPES_H)
 
-configure_file(expat_config.h.cmake expat_config.h)
+configure_file(${CMAKE_CURRENT_LIST_DIR}/expat_config.h.cmake expat_config.h)
 add_definitions(-DHAVE_EXPAT_CONFIG_H)
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -1,0 +1,119 @@
+# CMAKE_BUILD_TYPE is required for PCHSupport.cmake
+if(NOT DEFINED CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE "Release" CACHE STRING
+    "Choose the type of build, options are: None(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) Debug Release RelWithDebInfo MinSizeRel.")
+endif()
+
+project(WinSparkle)
+
+cmake_minimum_required(VERSION 2.8.12)
+
+if(NOT MINGW)
+  message(FATAL_ERROR "Can be used only with MinGW")
+endif()
+
+cmake_policy(SET CMP0022 NEW)
+
+get_filename_component(ROOT_DIR ${CMAKE_SOURCE_DIR}/.. REALPATH)
+
+# Parse version header to always have actual version of winsparkle
+file(READ "${ROOT_DIR}/include/winsparkle-version.h" VERSION_FILE)
+string(REGEX REPLACE "^.*#[ \t]*define[ \t]+WIN_SPARKLE_VERSION_MAJOR[ \t]+([0-9]+).*$" "\\1" LIB_MAJOR_VERSION ${VERSION_FILE})
+string(REGEX REPLACE "^.*#[ \t]*define[ \t]+WIN_SPARKLE_VERSION_MINOR[ \t]+([0-9]+).*$" "\\1" LIB_MINOR_VERSION ${VERSION_FILE})
+string(REGEX REPLACE "^.*#[ \t]*define[ \t]+WIN_SPARKLE_VERSION_MICRO[ \t]+([0-9]+).*$" "\\1" LIB_PATCH_VERSION ${VERSION_FILE})
+
+set(LIB_VERSION_STRING "${LIB_MAJOR_VERSION}.${LIB_MINOR_VERSION}.${LIB_PATCH_VERSION}")
+
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin" )
+# Use the same path for shared and static library
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib" )
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib" )
+
+set(BIN_INSTALL_DIR "bin")
+set(LIB_INSTALL_DIR "lib")
+set(INCLUDE_INSTALL_DIR "include")
+set(CONFIG_INSTALL_DIR "${LIB_INSTALL_DIR}/cmake/${PROJECT_NAME}")
+set(CMAKE_DEBUG_POSTFIX "d")
+
+set(EXPAT_INCLUDE_DIRS "${ROOT_DIR}/3rdparty/expat/lib")
+
+# bundled expat
+add_subdirectory(expat)
+
+# bundled wxWidgets
+add_subdirectory(wxWidgets)
+
+# WinSparkle
+set(SOURCE_DIR ${ROOT_DIR}/src)
+
+include_directories(${ROOT_DIR}/include)
+include_directories(${EXPAT_INCLUDE_DIRS})
+include_directories(${wxWidgets_INCLUDE_DIRS})
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+
+add_definitions(
+  -DWINVER=0x0600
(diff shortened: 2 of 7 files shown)


vslavik:
I’ll start by saying that having MinGW support would be nice — even though WinSparkle is buildable with free version of VS, being able to build with gcc would be welcome.

However, I take issue with the use of CMake. Using it as the only build system is out of the question (due to its requirement that all users have it installed) and maintaining two parallel build systems would be wasteful. The projects in there are already generated using [Bakefile](http://bakefile.org) and a better approach would arguably be to make its GNU makefiles backend work with MinGW, which should be pretty minor. I’m not closing this outright, but I’m not sure whether to merge it.

At the very least, it’s my understanding (see #62) that this doesn’t even work — and so is unmergeable by definition. That aside, it would be easier to live with if the CMake files weren’t all over the place and if it didn’t significantly change the imported expat code (putting one dependency in `3rdparty/CMakeLists.txt` and the other in `3rdparty/something/CMakeLists.txt` in particular makes little sense).

drizt:
I didn't work with Bakefile. I looked at Bakefile documentaion but didn't see MinGW support. If Bakefile supports MinGW probably using of Bakefile will be the best approach.

drizt:
I added `WX_PRECOMP` to defines. It solved compilation problem.

vslavik:
> I added WX_PRECOMP to defines. It solved compilation problem.

Congratulations on making the build significantly slower — you really should *enable precompiled headers* as I wrote above...

drizt:
Maybe you right. I never used precompiled headers before. It's not applied in gcc world.

drizt:
Added procompiled headers support

drizt:
Can't compile with latest changes. I got such error

    [ 70%] Building CXX object 3rdparty/CMakeFiles/wxWidgets.dir/wxWidgets/src/msw/dir.cpp.obj                                                                                                                                                   
    /home/taurus/develop/winsparkle/3rdparty/wxWidgets/src/msw/dir.cpp: In member function 'bool wxDirData::Read(wxString*)':
    /home/taurus/develop/winsparkle/3rdparty/wxWidgets/src/msw/dir.cpp:265:44: error: call of overloaded 'wxLogSysError(DWORD&, const wxString&, wxCStrData)' is ambiguous
                               m_dirname.c_str());
                                            ^
    /home/taurus/develop/winsparkle/3rdparty/wxWidgets/src/msw/dir.cpp:265:44: note: candidates are:
    In file included from /home/taurus/develop/winsparkle/3rdparty/wxWidgets/include/wx/string.h:46:0,
                     from /home/taurus/develop/winsparkle/3rdparty/wxWidgets/include/wx/log.h:43,
                     from /home/taurus/develop/winsparkle/3rdparty/wxWidgets/include/wx/msw/private.h:23,
                     from /home/taurus/develop/winsparkle/build/3rdparty/wxprec.h:44:
    /home/taurus/develop/winsparkle/3rdparty/wxWidgets/include/wx/log.h:1489:31: note: void wxLogSysError(long int, const wxFormatString&, T1) [with T1 = wxCStrData]
         WX_DEFINE_VARARG_FUNC_NOP(wxLog##level, 2, (argclass, const wxFormatString&)) \
                                   ^
    /home/taurus/develop/winsparkle/3rdparty/wxWidgets/include/wx/strvararg.h:1156:10: note: in definition of macro '_WX_VARARG_DEFINE_FUNC_NOP'
         void name(_WX_VARARG_FIXED_UNUSED_EXPAND(numfixed, fixed),                \

Seems problem in `wxUSE_INTL` now is `1`. 

Here 

            wxLogSysError(err, _("Cannot enumerate files in directory '%s'"), m_dirname.c_str());

`_("Cannot enumerate files in directory '%s'")` has type wxString but `m_dirname.c_str()` is not. If use such argument ``_("Cannot enumerate files in directory '%s'")`.c_str()` it will be compiled fine.

drizt:
Seems I found an error. I trying to compile wxWidgets (without WinSparkle) on Fedora 22. So.

    $ mingw32-configure --disable-log
    ...
    $ make -j5
    ...
    ./src/msw/dir.cpp: In member function 'bool wxDirData::Read(wxString*)':
    ./src/msw/dir.cpp:265:66: error: call of overloaded 'wxLogSysError(DWORD&, const wxString&, wxCStrData)' is ambiguous
                                                 m_dirname.c_str());


With `--enable-log` no error.

drizt:
Fill bugreport on trac.wxwidgets.org

vslavik:
> Fill bugreport on trac.wxwidgets.org

Please don't. This is well documented in wx's docs.

drizt:
Too late. http://trac.wxwidgets.org/ticket/17048. Anyway in documentation I can't find topic "How to build wxWidgets.". Also any compilation error on `make` (after `./configuration`) it's not OK IMHO.

vslavik:
Ah, you're talking about building wx itself. No idea why you add such irrelevant things here, then :(

drizt:
Then remove this noise.

drizt:
commit [b08ca17749ba3cdd86f83e1980a25680ee02d968](https://github.com/wxWidgets/wxWidgets/commit/b08ca17749ba3cdd86f83e1980a25680ee02d968) in wxWidgets submodule fixed compilation error. I tried to build with this commit. Now it works. Please update wxWidgets submodule.

drizt:
I noticed you updated wxWidgets. I rebased my cmake branch to latest winsparkle source. Now all works. Can we continue this PR?

vslavik:
> Can we continue this PR?

Please see my very first comment. Make it so that it is easy to update the Cmake files and so that they stay largely out of the way and can be ignored and I’m inclined to merge it then. Preferably be means of generating them from the same source in the first place (which I have a feeling would be less work than this), but failing that, at least by organizing them better instead of leaving these huge files full of custom code scattered inconsistently all over the place (see above).  Ideally by putting it all under `cmake/` so that I can largely ignore its existence.

GitCop:

 * Commit: b9804352c4cada83a11a5518a85ff4e0202e025d

---

drizt:
All CMake stuff moved to cmake subfolder.
Notice I edited *3rdparty/expat/ConfigureChecks.cmake*. Without such changing I can't use *ConfigureChecks.cmake* outside the *3rdparty/expat* folder.

GitCop:

 * Commit: b9804352c4cada83a11a5518a85ff4e0202e025d

---

vslavik:
> Without such changing I can't use ConfigureChecks.cmake outside the 3rdparty/expat folder.

You shouldn’t have any need or reason to use it at all in the first place: you’re not targeting all platforms Expat supports, just a single one, compiling on which is trivial, as you can observe in `expat.bkl`.

Please also see https://github.com/agis-/git-style-guide#commits about commit messages and organization, length limit is the least of it.

drizt:
> You shouldn’t have any need or reason to use it at all in the first place: you’re not targeting all platforms 
MinGW is not Visual Studio. `expat.bkl` defines `COMPILED_FROM_DSP` which suitable for VS based projects. I can't be sure that is correct for MinGW furthermore for **any** version of MinGW. Or if/when expat will be updated meaning of `COMPILED_FROM_DSP` can be changed. `ConfigureChecks.cmake` checks environment and set defines for current system. So in this case I can be sure that it is ok. Unfortunately in this case small fix of `ConfigureChecks.cmake` is required.

vslavik:
> I can't be sure that is correct for MinGW furthermore for any version of MinGW.

Changes are it is (given what CRT MinGW uses), but OK, that’s a fair point…

> Or if/when expat will be updated

That’s a good one. Hell will probably freeze over before that happens, even if I don’t remove this dependency sooner than that...

drizt:
Updated commit messages. Now

drizt:
ping

vslavik:
> ping

FYI, the only thing you accomplish with bugging is more delays. Just don’t.

If you want this merged faster, why don’t you, instead of bugging, focus on improving it. See any git guide on the commit messages: the idea is not to fill a paragraph, but to put *useful and informative* description there. Talking about some future commits or rationale for splitting commits (i.e. git 101) is totally off-topic, so is stating the obvious that is visible from the commit itself (“all placed in cmake directory”) as opposed to explaining things (“put everything in cmake directory to interfere with the official build as little as possible”). 

On to the diff itself: I very much doubt there’s any need for that `${CUR_DIR}` everywhere (which, to make it worse, has value *that is not “current dir”* as it calls itself!). And I don’t think it makes sense to mirror the source directories tree under `cmake` — why?! Make it clean, simple and easy to use. If that requires doing things differently, argue your case. But this is a mess, there’s no rational reason to have e.g. `PCHSupport.cmake` under `3rdparty`, short of it being the simplest thing you could do.

drizt:
Improved commit messages.

Reorganised files in cmake folder. Some folders is required becouse each CMake project must use own directory and own CMakeLists.txt file. But I reduced files hell.

Also now `CMAKE_BUILD_TYPE=Release` by default.

vslavik:
Sorry, I apparently forgot to push my local copy :( Done now. Thanks!

Issue #35, opened 2014-10-24
support multilingual releaseNotesLink (xml:lang)


kenjiuno:
Hi.

I'm willing to prepare at least 2 releaseNotes for English and Japanese within single appcast xml.

From a certain sparkle documentation, at Localization section, it says xml:lang is useful for localization.
https://github.com/sparkle-project/Sparkle/wiki/Publishing-An-Update

If I tried that, the some releaseNotes URLs were concatenated into one URL.

Is there any chance to support xml:lang attribute that is seen in that documentation so that we can supply multilingual releaseNotes from single appcast xml?

PR #65, opened 2015-05-28
Localization Support

ManicMeerkat:
I needed to localize WinSparkle into 8 languages (with more on the way). The project was set up such that the main application was written in C++/CLI with all localized strings come from a .NET resource DLL. In order to provide as much flexibility for translations and keep all the localized strings in the same place to maintain the localization workflow, I added the following method to the WinSparkle API:

void win_sparkle_set_localized_string(WinSparkleStringID stringID, const wchar_t* localized)

where WinSparkleStringID is an enum found in the new stringids.h, and which corresponds to a UI element.

The client will call win_sparkle_set_localized_string() for each WinSparkleStringID for which it has a translation. If the client doesn't provide a translation, the default English string will be used.


vslavik:
> The client will call win_sparkle_set_localized_string() for each WinSparkleStringID for which it has a translation.

WinSparkle uses wxWidgets, which already has a *sane* i18n interface — gettext (see also #35). Any solution that uses *string tables* of all things, requires tons of hardcoded API calls to set strings, doesn’t bundle translations with WinSparkle itself isn’t acceptable. This is just not a good approach from the technical POV.

I won’t be merging this, sorry.

PR #75, opened 2015-10-07
Visual Studio 2015 support

statico:
This pull request adds project files for MSVS 2015 and updates the wxWidgets dependency.

However, likely won't compile without https://github.com/wxWidgets/wxWidgets/pull/111


GitCop:

 * Commit: c9e5de8989d5639b2f8d31b1d3138a482adedd09

---

vslavik:
The majority of this PR is unrelated to VS2015.

You misunderstand what the deal with `wxMSWDisableSettingHighDPIAware()` is — it is not that it "appears to no longer be available" (wx really doesn't have that loose approach to backward compatibility), it is that it is [explicitly documented to be 3.0 only](https://github.com/wxWidgets/wxWidgets/blob/WX_3_0_BRANCH/interface/wx/init.h#L111) because wx 3.0 *did unconditionally opt-in HiDPI*. And the only reason you had to deal with it is that you updated wx (without any in-commit explanation of why, by the way — so: why?) to the in-development `master` branch rather than to the latest stable series (`WX_3_0_BRANCH`) that WinSparkle uses and should continue to use.

So those bits are undesirable, which leaves 647c994. The only issue I have with that is that it adds these files manually. If you look at the `vcxproj` files (even the ones you added!), they have a header saying they were generated. I have no particular desire to maintain some files manually and have some generated, so I'd rather not do it like this. There are two "proper" ways to add 2015 projects:

1. Add 2015 support to [Bakefile](http://bakefile.org), which should be trivial; or
2. Merge all vcxproj files into one that is multi-targeting VS versions from 2010 onwards, remove support for 2008, and continue to maintain vcxproj manually instead of relying on Bakefile. [This is possible with MSBuild](https://github.com/wxWidgets/wxWidgets/blob/master/build/msw/wx_config.props) and I think preferred (it helps with #73 too).

If you'd be willing to address that, it would be appreciated, otherwise I'm OK with merging this as a temporary solution, but *without* the upgrade-wx-to-master bit. It looks to me like the `WX_3_0_BRANCH` branch has 2015 support now too.

statico:
Thanks @vslavik. Let me preface by saying that I have no idea what I'm doing.

Regarding the wxWidgets upgrade, oops! That was a red herring so I've reverted that. There will be `Please update wx/compiler.h to recognize this VC++ version` messages while building, but they're just warnings.

I had no idea this project used a build system -- I've never seen Bakefile before. I'm afraid that I don't have the time to contribute to Bakefile, nor am I proficient enough to make a multi-targeted VS project. Feel free to merge or close at will.

vslavik:
>  Let me preface by saying that I have no idea what I'm doing.

That’s OK, that’s what code reviews are *for* ;)

> There will be `Please update wx/compiler.h to recognize this VC++ version` messages while building with VS2015

That’s why I wrote about that you should instead update wxWidgets to the tip of the `WX_3_0_BRANCH` (which recognizes VS2015), *not* that you shouldn’t update it at all...

vslavik:
Merged in e68bafd7 + 179d4107.

PR #77, opened 2015-10-14
Check for updates with UI and install automatically

statico:
This PR adds the ability for a program to check for updates and automatically install them if available. This functionality is present in the Mac version of Sparkle as of https://github.com/sparkle-project/Sparkle/pull/620.

Once `win_sparkle_check_update_with_ui_and_install()` is called, the following cases may occur:

- An error occurs, in which case the `win_sparkle_error_callback_t` is called
- The application is up to date, in which case the `win_sparkle_did_not_find_update_callback_t` is called
- The application is not up to date. The update begins to download automatically, then...
   - If the user interrupts the download (i.e.., clicks the cancel button or closes the update dialog) the `win_sparkle_update_cancelled_callback_t` is called
   - If the update completes, `win_sparkle_shutdown_request_callback_t` is called and the newly-downloaded installer runs

CC @mlogan @BKcore

### DIFF
--- a/include/winsparkle.h
+++ b/include/winsparkle.h
@@ -266,6 +266,16 @@ WIN_SPARKLE_API int __cdecl win_sparkle_get_update_check_interval();
 */
 WIN_SPARKLE_API time_t __cdecl win_sparkle_get_last_check_time();
 
+/// Callback type for win_sparkle_error_callback()
+typedef void (__cdecl *win_sparkle_error_callback_t)();
+
+/**
+    Set callback to be called when the updater encounters an error.
+
+    @since 0.5
+*/
+WIN_SPARKLE_API void __cdecl win_sparkle_set_error_callback(win_sparkle_error_callback_t callback);
+
 /// Callback type for win_sparkle_can_shutdown_callback()
 typedef int (__cdecl *win_sparkle_can_shutdown_callback_t)();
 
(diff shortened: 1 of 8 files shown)


GitCop:

 * Commit: 41dd1c2a8859db593207cb88e3d1bc93a568135b

---

vslavik:
Thanks!

I’ll have some code-level comments too, but at the high-level:

> An error occurs, in which case the win_sparkle_update_cancelled_callback_t is called

It doesn’t seem correct to call something called “cancelled” on *error*, instead of when *the user* cancels the update… Sounds to me like a dedicated error callback would make more sense.

More generally, I question the entire purpose of these callbacks (with the exception of reporting errors, which makes sense) — what are they useful for? You’re vaguely alluding to their usefulness, but I’d like to see some *specific* example of what they are good for, i.e. what the already existing callbacks can’t do. I’d also say something is wrong when you find yourself documentation a function (`win_sparkle_set_did_find_update_callback()`) with basically “this is not useful”… 

Anyway, I assume there *is* a good reason for those, I’d just like to understand it better. It also seems this is not directly related to `win_sparkle_check_update_with_ui_and_install`, could you move it into a separate commit (see the gitcop-linked guide, one commit per one logical thing)? And add more explanatory descriptions to both commits (ditto)? Big changes like this deserve it and "Add win_sparkle_check_update_with_ui_and_install” is a bit too brief when digging through the history later…

vslavik:
Globally, I like this patch a lot (the comments above are really just trivia) — and it’s great to see it took relatively little code. Thanks!

statico:
Thanks, I'll address these comments shortly.

> More generally, I question the entire purpose of these callbacks (with the exception of reporting errors, which makes sense) — what are they useful for? You’re vaguely alluding to their usefulness, but I’d like to see some specific example of what they are good for...

As I mentioned in the Mac Sparkle PR, we're using Sparkle in a wrapper application whose purpose is solely to check for updates, respond to custom URI schemes, and launch the main application which is a giant cross-platform blob. The blob (a game, actually) fetches remote resources, and as we push new resources they become backward incompatible with older app versions, so it's imperative that we keep the application up to date.

Currently, the wrapper application looks like this:
2015-10-14 10_54_57-native-client running - microsoft visual studio administrator

These callbacks allow us to skip this dialog entirely -- we can now check for updates, install it if one is available, or run the game if one isn't. The callback approach seemed to be the most straightforward way to implement this functionality. When an update is available, it installs and the application restarts. When no update is available, the game launches. When the update is cancelled, the application quits.

GitCop:

 * Commit: b49789fb8aae2eef6fd8e10dd16bbf842ed70272

---

statico:
I think I've addressed your comments from https://github.com/artillery/winsparkle/commit/ac580c312973f81dc6be040892f5654b796ec43d. Please take another look.

statico:
Also, I updated the initial PR comment to be accurate.

vslavik:
Merged, thank you!

PR #82, opened 2015-12-10
Always make call into "did not find update" callback

firesnatch:
Updates to ui.cpp

Always make call into "did not find update" callback

Previously, win_sparkle_set_did_not_find_update_callback would not get called if win_sparkle_check_update_without_ui() was used.  Made changes so that callback would get called when win_sparkle_check_update_with_ui_and_install(), win_sparkle_check_update_without_ui(), etc are used.

### DIFF
--- a/src/ui.cpp
+++ b/src/ui.cpp
@@ -739,9 +739,7 @@ void UpdateDialog::StateNoUpdateFound(bool installAutomatically)
 {
     m_installAutomatically = installAutomatically;
 
-    ApplicationController::NotifyUpdateNotFound();
-
-    if ( m_installAutomatically )
+       if ( m_installAutomatically )
     {
         Close();
         return;
@@ -1485,6 +1483,7 @@ void UI::NotifyNoUpdates(bool installAutomatically)
     EventPayload payload;
     payload.installAutomatically = installAutomatically;
 
+    ApplicationController::NotifyUpdateNotFound();
     if ( !uit.IsRunning() )
         return;
 


GitCop:

 * Commit: 4b89e9a62e65c9a744de3afc9749367bfc74bbe9

---

vslavik:
Are other callbacks also affected, or just this one?

firesnatch:
When win_sparkle_check_update_with_ui() is used, all of the callbacks
work.  However, when win_sparkle_check_update_without_ui() is used:

win_sparkle_set_update_cancelled_callback - WORKS
win_sparkle_set_shutdown_request_callback - WORKS
win_sparkle_set_did_not_find_update_callback - DOES NOT WORK.

So I believe win_sparkle_set_did_not_find_update_callback is the only one
with a problem.  This is what my commit fixes.

I have not tested win_sparkle_check_update_with_ui_and_install();

Also I noticed that if the appcast.xml has bad sparkle:installerArguments
in the enclosure tag, you get a "Failed to launch the installer" message,
and none of the callbacks get called.  This happens with both
win_sparkle_check_update_with_ui() and
win_sparkle_check_update_without_ui().  I have not attempted to fix this in
the code.  It can be worked around by using correct arguments in the
appcast.xml.

Hope that helps.

- Sean

vslavik:
Thanks! What about `win_sparkle_set_error_callback`?

The `update_cancelled` and `shutdown_request` are unrelated, they coordinate hosting app shutdown (unlike the others that inform it about what’s happening). It’s expected that they don’t get called if installation doesn’t run, because in that case, there’s no intention to restart the hosting app.

firesnatch:
I'm not sure how to test  win_sparkle_set_error_callback.  How can I
trigger that callback?  Let me know, and I'll test it out.

vslavik:
> I'm not sure how to test  win_sparkle_set_error_callback.  How can I
trigger that callback? 

You could e.g. add ` throw std::runtime_error("dummy");` to [`UpdateDownloader::Run`](https://github.com/vslavik/winsparkle/blob/0c860b728cf5268872a18ab23b018cac98732100/src/updatedownloader.cpp#L167) to simulate this.

Looking at the code in `UI::NotifyUpdateError()`, I think it needs the same fix you applied to the other callback, but I’m not 100% sure.

firesnatch:
It doesn't look like a change is needed for that.  I triggered the error
like you said, and every time the error callback was called.  I tested with:

win_sparkle_check_update_without_ui()
win_sparkle_check_update_with_ui_and_install();
win_sparkle_check_update_with_ui();

- Sean

vslavik:
Sorry, I mislead you about how to test the error case — that was too late, when UI is shown, so yes, there was no difference. I just checked and error callback suffers from the same, but let’s leave that for another day: both that one and the update-cancelled ones are apparently intentionally done after the UI is closed (which I don’t think is best, but I’m not ready to address it right now).

This PR committed as 9e4eaeb8 (after fixing spurious whitespace changes and commit message).

firesnatch:
Cool, thanks.  Sorry about the whitespace, it was my first time using the
github web site to input the changes needed.  I've only used the command
line tool on linux.

PR #84, opened 2016-01-25
Added callback for exeption logging

Houd1ny:


vslavik:
Your code is messily confusing errors and exceptions – pick one and stick to it (and don't pick "exceptions" because they clearly are not).

More importantly, you seem to be oblivious to the existence of the same thing (as far as I can tell) on the master already. That's contributing 101: the first thing to do is to check the current code.

Houd1ny:
@vslavik  Hi. I know there are already callback(win_sparkle_set_error_callback). But it only handle ui errors. But you can`t get error from wxWingets in case of crash there. About it`s name I agree it not handle exception It just logs it text .

vslavik:
OK, then that explanation should have been part of the commit message (see https://github.com/agis-/git-style-guide) or at least the PR’s description.

But I still remain confused by this PR. I don’t think the existing code handles “UI errors” only (what is that, anyway? I don’t think _any_ errors in WinSparkle are “UI” ones). And I certainly don’t know why are you bringing _crashes_ up — that’s neither errors, nor exceptions, and your code certainly doesn’t handle those.

So: if you think the current code doesn’t handle callbacks for something and it should, please reopen with updated commit, that explains itself and that isn’t internally confused about the meaning of error/exception/crash. It would also be nice if you could provide a _specific_ example (e.g. in the form of a feed URL) that should be reported, but isn’t, and provide some explanation of _why_ it should be handled by the callback.

Houd1ny:
My aim was to added user defined logger callback. Right now all errors in showing using  OutputDebugStringA. This is good for development on local machine. But it also possible that customer wants to track amount and types error for users by sending that data to some online service or write in some log file. It is not implemented right now.

Houd1ny:
I agree there should be only one error callback or this will be mess

vslavik:
You need to work on your terminology then — that’s neither errors or “UI errors", nor exceptions, nor crashes, that’s *logging*. 

FWIW, I think it’s a bit pointless to have a logging framework included when you can [capture debug output](http://stackoverflow.com/questions/24320665/log-outputdebugstring-to-a-file-without-debugview).

PR #85, opened 2016-01-25
Make installerArguments work with MSI.

Marginal:
The [documentation](https://github.com/vslavik/winsparkle/wiki/Appcast-Feeds#installer-arguments) says that you can use <sparkle:installerArguments>="/passive" when installing an MSI file. But this doesn't actually work.

WinSparkle uses wxExecute() instead of wxLaunchDefaultApplication() when installer arguments are supplied in the appcast. But wxExecute() ultimately calls win32's CreateProcess(), which expects an executable not a MSI file.

This patch copies wxLaunchDefaultApplication() to use win32's ShellExecuteEx(), which handles both executables and MSI files.

### DIFF
--- a/src/ui.cpp
+++ b/src/ui.cpp
@@ -57,6 +57,7 @@
 #include <exdisp.h>
 #include <mshtml.h>
 #include <commctrl.h>
+#include <shellapi.h>
 
 
 #if !wxCHECK_VERSION(2,9,0)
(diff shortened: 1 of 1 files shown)


vslavik:
Thanks! This is the best kind of patch: it fixes a bug *and* simplifies the code.

Could you please make the cosmetic updates to c4a56f1 and force-push updated version? TIA!

Marginal:
I've removed those two comments.
Happy to force-push to consolidate. But first are you happy with ec1edc4 ?

vslavik:
Wow, that was fast. Yep, thanks!

Marginal:
Done

vslavik:
Thanks!

PR #99, opened 2016-04-26
Pascal bindings

teo-tsirpanis:
File winsparkle.h is ported to Pascal.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ translations/*.mo
 3rdparty/expat/win32/bin
 3rdparty/expat/win32/tmp
 packages
+pascal/lib/
+pascal/backup/
--- a/pascal/winsparkle.pas
+++ b/pascal/winsparkle.pas
@@ -1,0 +1,258 @@
+{
+   *  This file is part of WinSparkle (https://winsparkle.org)
+   *
+   *  Copyright (C) 2009-2016 Theodore Tsirpanis
+   *
+   *  Permission is hereby granted, free of charge, to any person obtaining a
+   *  copy of this software and associated documentation files (the "Software"),
+   *  to deal in the Software without restriction, including without limitation
+   *  the rights to use, copy, modify, merge, publish, distribute, sublicense,
+   *  and/or sell copies of the Software, and to permit persons to whom the
+   *  Software is furnished to do so, subject to the following conditions:
+   *
+   *  The above copyright notice and this permission notice shall be included in
+   *  all copies or substantial portions of the Software.
+   *
+   *  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+   *  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+   *  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+   *  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+   *  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+   *  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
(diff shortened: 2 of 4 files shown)


vslavik:
Wow, this is surprisingly simple.

Can you clean it up a bit before I commit it, though (and force-push tidied commits)? 

* This really is one change, not two, so should be one commit — see also https://github.com/agis-/git-style-guide#commits
* Please put it all into a `pascal` directory, ideally with any instructions that may be needed.
* I’m going to add a few inline comments to the code for minor things as well.

winsparkle.pas
```diff
@@ -0,0 +1,413 @@
...
+   *  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+   *  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+   *  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+   *  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+   *  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+   *  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+   *  DEALINGS IN THE SOFTWARE.
+   *
+ }
+//Converted to Pascal by Theodore Tsirpanis.
```
vslavik: This wasn’t really “converted”. I think it would be appropriate to put yourself into the Copyright line above instead of this comment.

winsparkle.pas
```diff
@@ -0,0 +1,413 @@
...
+{$FATAL WinSparkle requires Windows}
+{$ENDIF}
+
+interface
+
+uses
+  ctypes;
+
+const
+  WinSparkleLib = 'winsparkle.dll';
```
vslavik: Should be `WinSparkle.dll` for consistency with everything else.

winsparkle.pas
```diff
@@ -0,0 +1,413 @@
...
+
+  {--------------------------------------------------------------------------*
+                         Initialization and shutdown
+   *-------------------------------------------------------------------------- }
+  {*
+      @name Initialization functions
+    }
+
+  {*
+      Starts WinSparkle.
```
vslavik: Is there any reason to keep these comments here? I.e. does the IDE use them? If not, then I’d prefer to remove them from this file (they are already in the C header — I’ll add a doxygen-generated copy to the site to make accessing the docs easier) and keep the file small, with just the declarations.

teo-tsirpanis:
OK. I am preparing a class-cased port of WinSparkle headers.

I have one question: Can I set the callbacks before I call `win_sparkle_init`?

vslavik:
Yes, you can. In fact, most initialization should be done before calling it. It’s an unfortunate name, isn’t it :( I wonder if it’s worth renaming to `win_sparkle_start()` or something…

teo-tsirpanis:
OK. I made it.

vslavik:
Thanks, and sorry for the delay! Could you however move the files to a `pascal/` subdirectory as I mentioned, instead of putting them into `include/`, which is for *C headers*? 

I also don’t understand what’s up with the `.gitignore` changes: surely something called “backup” has no place there, does it? And is a “lib” subdirectory special in some way? I.e. is something (what, given the lack of any project files or such?) always put there, regardless how you include the `winsparkle.pas` file in your code? I rather suspect it’s not, but then it’s been ~20 years since I last used any variant of Pascal myself…

GitCop:

 * Commit: 365d0375892a4ab1a91c9938692101e397358e25

---

teo-tsirpanis:
The Lazarus IDE by default backups edited files in a `backup` subdirectory. Additionally, the included Lazarus package places the compiled binaries in a `lib` subdirectory.

vslavik:
> The Lazarus IDE by default backups edited files in a backup subdirectory.

Wow, that’s… weird.

> Additionally, the included Lazarus package places the compiled binaries in a lib subdirectory.

Sorry, I completely missed that file when reviewing the diff :-(

Thanks a lot for updating the PR — and of course contributing this in the first place! Merged now.

Issue #109, opened 2016-07-12
Winsparkle won't follow redirects


ThanosSiopoudis:
If the item URL in the appcast xml file points to a redirect location (to prevent hotlinking for example), WinSparkle fails when attempting to download the file, as it won't follow the LOCATION header.

The runtime_error thrown is "Cannot save update file"

vslavik:
Please be a good developer and reports bugs as you’d want bugs reported to you. *Always* include [steps to reproduce](http://www.chiark.greenend.org.uk/~sgtatham/bugs.html).

At the barest minimum, reproduction steps must include specifics, i.e. the URL of the feed that doesn’t work and preferably (it’s dead simple, so there’s no excuse to not do it, really) a patch against the included example that reproduces the issue.

WinInet handles redirects automatically be default. It would be very strange indeed if that didn’t work in WinSparkle, no least *because I use them myself*.

ThanosSiopoudis:
Thanks, I'm trying to gather all the data before I can help you with what's going on. Bear with me, it's my first crack at Winsparkle, and there's a lot of code to go through.

It seems that `InternetOpenUrlA` is not the issue, as it DOES follow the redirect and actually downloads the file data.

The problem seems to be in the method that sets the filename, and not actually downloading the file.

Here's an appcast I've created that doesn't work, feel free to try it, and let me know if you don't have access to it:
https://rink.hockeyapp.net/api/2/apps/132dac157c62466b990de9d58ac36427.rss

PR #110, opened 2016-07-12
Get the file URL from the response

ThanosSiopoudis:
This should fix the issue in #109 

The provided PR will attempt to get the URL from `INTERNET_OPTION_URL` by querying the response. This should prevent issues where the URL in an AppCast file points to a location which is being redirected.

### DIFF
--- a/src/download.cpp
+++ b/src/download.cpp
@@ -99,10 +99,10 @@ bool GetHttpHeader(HINTERNET handle, DWORD whatToGet, DWORD& output)
            ) == TRUE;
 }
 
-std::wstring GetURLFileName(const URL_COMPONENTSA& urlc)
+std::wstring GetURLFileName(const char *url)
 {
-    const char *lastSlash = strrchr(urlc.lpszUrlPath, '/');
-    const std::string fn(lastSlash ? lastSlash + 1 : urlc.lpszUrlPath);
+    const char *lastSlash = strrchr(url, '/');
+    const std::string fn(lastSlash ? lastSlash + 1 : url);
     return AnsiToWide(fn);
 }
 
@@ -206,7 +206,25 @@ void DownloadFile(const std::string& url, IDownloadSink *sink, int flags)
 
     if ( !filename_set )
     {
-        sink->SetFilename(GetURLFileName(urlc));
+        char *optionurl;
+        DWORD ousize = 0;
+        InternetQueryOptionA(conn, INTERNET_OPTION_URL, nullptr, &ousize);
+        if (GetLastError() == ERROR_INSUFFICIENT_BUFFER)
+        {
+            optionurl = new char[ousize];
+            if ( InternetQueryOptionA(conn, INTERNET_OPTION_URL, optionurl, &ousize) )
+                sink->SetFilename(GetURLFileName(optionurl));
+            else
+                sink->SetFilename(GetURLFileName(urlc.lpszUrlPath));
+            
+            delete[] optionurl;
+        }
+        else
+        {
+            sink->SetFilename(GetURLFileName(urlc.lpszUrlPath));
+        }
+        
+        optionurl = NULL;
     }
 
     // Download the data:


vslavik:
Thanks! I appreciate the PR, but there are some small issues that it would be great if you could fix (and replace the current commit with `git push -f`):

src/download.cpp
```diff
@@ -206,6 +206,13 @@ void DownloadFile(const std::string& url, IDownloadSink *sink, int flags)
 
     if ( !filename_set )
     {
+        char ioptionurl[512];
```
vslavik: If you want to use a fixed buffer, 2048 would be a better size, because it’s the IE/WinInet URL length limit, unlike 512, which is arbitrary. (Getting the length dynamically)[https://github.com/SortByte/Network-Manager/blob/2c80507d3a8f8c73d68df94baf9a0e6a0f698147/Launcher/Network.cpp#L169] would probably be even better, though.

src/download.cpp
```diff
@@ -206,6 +206,13 @@ void DownloadFile(const std::string& url, IDownloadSink *sink, int flags)
...
+        DWORD ousize = 512;
```
vslavik: In the static buffer case, this should be `= sizeof(ioptionurl)` — it’s safer. Why ‘ousize’, BTW? Was that meant to be ‘outsize’?

src/download.cpp
```diff
@@ -206,6 +206,13 @@ void DownloadFile(const std::string& url, IDownloadSink *sink, int flags)
...
+        if (InternetQueryOptionA(conn, INTERNET_OPTION_URL, ioptionurl, &ousize))
+        {
+            urlc.lpszUrlPath = ioptionurl;
```
vslavik: This seems rather hacky. Why not change `GetURLFileName` to take the URL string as its argument and call it as `GetURLFileName(ioptionurl)` (here) or `GetURLFileName(urlc.lpszUrlPath)` (below in else branch) respectively?

src/download.cpp
ThanosSiopoudis: `ousize` from `optionurl`. Not great, I know.

I noticed I also omitted the leading and trailing space in the 'if' statement

src/download.cpp
ThanosSiopoudis: I used 512 as this is what is used a few lines above for the `url_path` buffer.
I'll change that to get it dynamically

vslavik: *blush* In that case it really doesn’t make sense to do any differently, sorry!

But in fact, there’s even simpler version of the code here, isn’t there? Since the URL is already being stored in the `url_path` buffer (which is where `urlc.lpszUrlPath` points to), what about just querying `INTERNET_OPTION_URL` into that directly and not having to allocate a second buffer:

    InternetQueryOptionA(conn, INTERNET_OPTION_URL, urlc.lpszUrlPath, &urlc.dwUrlPathLength);

? Am I missing something again?

ThanosSiopoudis: I used a different buffer to make sure there is a backup url to use in case InternetQueryOptionA fails somehow and ends up emptying the original buffer, leaving the app without a url to try at all. Surely this should never happen, but just covering all bases.
In any case either works for me, I've already changed it to use a dynamic size buffer, but I'm happy to change it to whatever you prefer.

vslavik: > but I'm happy to change it to whatever you prefer.

I’ll have you make the call — it’s obvious you have the choices well reasoned. However, *this* version of code for dynamically allocated buffer is buggy (it doesn’t check the first `InternetQueryOptionA` call’s result and it leaks memory b/c it never frees `optionurl`).

In any case, could you merge the final version into a single commit the describes the *why* (i.e. how it fixes the bug with failing to recognize redirected enclosure’s filename) rather than the *how* (which the code explains)? As it is now, it’s not clear from the commit message(s) alone *why* the changes are done (and the 25ee1d8 commit itself breaks compilation, breaking `git bisect`).

TIA!

ThanosSiopoudis: Will do.

On a side note, calling `InternetQueryOptionA` with a nullptr for buffer, makes it return `false` and that's why there's no check. It does however set the size ptr.

Do you prefer a fresh PR with a single commit?

ThanosSiopoudis: I ended up doing a rebase and squashing the commits into one. I'll fix the issues now and `git push -f`

GitCop:

 * Commit: 048fb0cb544f5a39fe8f03dc75affc7e4155e919

---

ThanosSiopoudis:
Ok, this should do it hopefully!

I've merged all commits into one, and fixed your points. As I said earlier, `InternetQueryOptionA` returns false when passed `nullptr` for buffer, but there is a specific message which is now checked with `GetLastError()` to determine if it failed for this reason.

I'm also clearing the buffer, and nullifying the pointer once done.

ThanosSiopoudis:
It seems that appveyor fails because it can't currently clone `git://git.code.sf.net/p/expat/code_git`, something wrong with it?

vslavik:
> It seems that appveyor fails because it can't currently clone git://git.code.sf.net/p/expat/code_git, something wrong with it?

Yes, SF being down.

vslavik:
Thanks! Rather than asking you for another iteration, I committed a slightly modified version as 6d0989db (under your name, of course) — it replaces needless resetting of `optionurl` with scoping, replaces `nullptr` with `NULL` (compatibility with older compilers) and adds an [explanatory commit message](http://chris.beams.io/posts/git-commit/).

ThanosSiopoudis:
Thanks, great to have this fixed in the master branch!

PR #113, opened 2016-08-31
support xml:lang in release notes

maxkadushkin:
following #35, implemented using xml:lang for multilingual releaseNotesLink. language isn't got from OS (need to do it?), but set up by win_sparkle_set_lang. 
the order of releaseNotesLink items doesn't matter. will be taken the first item without xml:lang tag and the last with suitable language


vslavik:
Thanks for your PR! I can’t merge it in its current state, unfortunately, could you please polish it per the inline comments I’ll add in a moment? I’d also appreciate better commit message with accurate title (#35’s one is more meaningful and understandable than the one you used; there’s nothing “dependent” there for one thing) and a description (the caveats about how it behaves present in this PR should be part of the commit message — see http://chris.beams.io/posts/git-commit/ for the rationale.)

src/appcast.cpp
```diff
@@ -256,7 +276,7 @@ void XMLCALL OnText(void *data, const char *s, int len)
     const size_t size = ctxt.items.size();
 
     if ( ctxt.in_relnotes )
-        ctxt.items[size-1].ReleaseNotesURL.append(s, len);
+        ctxt.items[size-1].ReleaseNotesURL.assign(s, len);
```
vslavik: I understand why you did this, but you need to avoid concatenating different links in another way. This is wrong, because Expat makes no guarantees about how often does it call this handler. It may decide to split any text block in multiple pieces, hence `append`. For example, it would be reasonable to clear `ReleaseNotesURL` when the tag is opened anew.

src/appcast.cpp
```diff
@@ -208,7 +228,7 @@ void XMLCALL OnEndElement(void *data, const char *name)
 
     if ( ctxt.in_item && strcmp(name, NODE_RELNOTES) == 0 )
     {
-        ctxt.in_relnotes--;
+        ctxt.in_relnotes && ctxt.in_relnotes--;
```
vslavik: Please don’t use oh-so-clever hacks like this, they only obfuscate the meaning and make it hard to read the code. Use `if` for branching in statements.

src/appcast.cpp
```diff
@@ -126,7 +130,23 @@ void XMLCALL OnStartElement(void *data, const char *name, const char **attrs)
     {
         if ( strcmp(name, NODE_RELNOTES) == 0 )
         {
-            ctxt.in_relnotes++;
```
vslavik: This (and the corresponding change when closing the tag) doesn’t seem right. `in_relnotes` keeps track of the position in the document, sometimes pretending the tag wasn’t there, as you now do, seems dangerous. Why not keep track of the name instead?

maxkadushkin:
yes, of cause, I'll examine your comments and my changes

vslavik:
> Merge with upstream

This makes merging the PR somewhat messy :( To reiterate, I'd be happy to merge this once the bugs are fixed.

maxkadushkin:
yes, of course. i have no PR without changes according your remarks

vslavik:
So much for your promise in 2016… or the spirit of contributing back :(

src/appcast.cpp
maxkadushkin: needs to skip unsuitable node. it was quick idea for that. and, I agree, not good idea. I modify it. I don't understand about the name. what the name?

maxkadushkin: xml looks like
<sparkle:releaseNotesLink>http://download.somesite.com/changes/changes.html</sparkle:releaseNotesLink>
<sparkle:releaseNotesLink xml:lang="de-DE">http://download.somesite.com/changes/changes_de.html</sparkle:releaseNotesLink>

PR #114, opened 2016-09-14
Detect download failures due to timeouts or network connectivity issues

nickdademo:
Detect download failures due to timeouts or network connectivity issues and throw an error (instead of erroneously showing the download as complete). Because these failures also return 'dwBufferLength = 0' like a successful download, we detect them by checking the number of received bytes against the Content Length (only if it is available). If there is a mismatch, an error is thrown. Otherwise, the download is considered successful.

### DIFF
--- a/src/download.cpp
+++ b/src/download.cpp
@@ -51,18 +51,24 @@ struct InetHandle
 
     ~InetHandle()
     {
-        if (m_handle)
-        {
-            if (m_callback)
-                InternetSetStatusCallback(m_handle, NULL);
-            InternetCloseHandle(m_handle);
-        }
+        Close();
     }
 
     void SetStatusCallback(INTERNET_STATUS_CALLBACK callback)
     {
         m_callback = callback;
         InternetSetStatusCallback(m_handle, m_callback);
+    }
+
+    void Close()
+    {
+        if (m_handle)
+        {
+            if (m_callback)
+                InternetSetStatusCallback(m_handle, NULL);
+            InternetCloseHandle(m_handle);
+            m_handle = NULL;
+        }
     }
 
     operator HINTERNET() const { return m_handle; }
@@ -141,6 +147,10 @@ void CALLBACK DownloadInternetStatusCallback(_In_ HINTERNET hInternet,
 
         case INTERNET_STATUS_REQUEST_COMPLETE:
             context->eventRequestComplete.Signal();
+            break;
+
+        case INTERNET_STATUS_CONNECTION_CLOSED:
+            context->conn->Close();
             break;
     }
 }
@@ -297,7 +307,14 @@ void DownloadFile(const std::string& url, IDownloadSink *sink, Thread *onThread,
         }
 
         if (ibuf.dwBufferLength == 0)
-            break; // all of the file was downloaded
+        {
+            // This check is required in case the INTERNET_STATUS_CONNECTION_CLOSED event was 
+            // received (and the handle was closed) during the call to InternetReadFileEx()
+            if (!conn)
+                throw Win32Exception();
+            else
+                break; // all of the file was downloaded
+        }
 
         sink->Add(ibuf.lpvBuffer, ibuf.dwBufferLength);
     }


GitCop:

 * Commit: 971e464d041d89b9cb88dd23ef8cee1dd733fdd7

---

vslavik:
In addition to GitCop’s complain (that’s one crazy *subject line*!), this doesn’t seem right. Errors are detect above, so I don’t see how could this happen due to an error or timeout — `GetError()` right above would report the error, wouldn’t it? In any case, you can’t just throw `Win32Exeption` when you don’t know if there *was* a win32 error reported — that’s the sort of things that leads to those unfamous “error: no error” messages.

Isn’t the *actual* bug that `WaitUntilSignaledWithTerminationCheck()` should be followed by some check for errors on `conn`? That would be more robust than trying to infer error condition from content length.

Do you have a reproducible test case?

vslavik:
>  `Detect download failures due to network issues.`

You missed the point of GitCop’s message. It’s not just the length, it’s about following the linked guidelines (which anyone using git should read). Removing all the useful explanations from your original commit message is *not* desirable: replacing it with a short summary subject (*without* the trailing dot), followed by empty line, followed by your very good description of what & why the commit does is what GitCop was asking for...

nickdademo:
The problem we are observing is that InternetReadFileEx() is returning true (even though the download has failed) and thus dwBufferLength is also zero. So checking the number of received bytes seemed to be a solution to differentiate this case from an actual successful download.

I think the easiest way to test this is to kill your internet connectivity during an installer download: the download will stall for around 10-20 seconds and then will report as completed. At least, this is what we have observed.

Also, I'll update the commit description to something more appropriate.

vslavik:
> The problem we are observing is that InternetReadFileEx() is returning true (even though the download has failed) and thus dwBufferLength is also zero. So checking the number of received bytes seemed to be a solution to differentiate this case from an actual successful download.

WinInet is a mature library. Jumping to the unsubstantiated conclusion that it is not properly reporting errors is near-certain to be wrong. By *far* more likely explanation is that my code *using* it is buggy. It’s not right to leave the bug(s) in and add a workaround on top of them, as this seems to be doing; better fix the underlying bug.

As I said, the code calls `InternetReadFileEx` and then *waits for async results to come in* (in `WaitUntilSignaledWithTerminationCheck`). If a problem happens during the wait, it’s not detected. So instead of counting bytes, that’s the place to check the status. I don’t know how off-hand and this is my first ever async use of WinInet (as the existence of this PR shows), but MSDN surely answers that.

> I think the easiest way to test this is to kill your internet connectivity during an installer download: the download will stall for around 10-20 seconds and then will report as completed. At least, this is what we have observed.

That’s of course how I was testing the code, so that’s not really reproducible way to demonstrate it...

nickdademo:
Actually, waiting in WaitUntilSignaledWithTerminationCheck() only occurs when InternetReadFileEx() returns FALSE with ERROR_IO_PENDING. In this case, the read is performed in the background and when complete, WaitUntilSignaledWithTerminationCheck() unblocks via the INTERNET_STATUS_REQUEST_COMPLETE event (a signal is sent from DownloadInternetStatusCallback). From what I've read, this is the correct usage of InternetReadFileEx() - so your code appears to be fine. However, the problem I am seeing is InternetReadFileEx() returning TRUE on a failed download for some reason - so there isn't any status information for me to get...

Anyway, I'll do some more investigation and get back to you. Hopefully there is a better way to do this.

nickdademo:
OK - it looks like I can do this more elegantly by handling the `INTERNET_STATUS_CONNECTION_CLOSED` event in `DownloadInternetStatusCallback`.

This is what is happening in the current (original) code:
1) Kill network connectivity: `InternetReadFileEx` fails with error `ERROR_IO_PENDING`
2) We now block and wait for the `INTERNET_STATUS_REQUEST_COMPLETE` event.

**[In between here we actually receive two events: `INTERNET_STATUS_CLOSING_CONNECTION` followed by `INTERNET_STATUS_CONNECTION_CLOSED` which are not handled by WinSparkle]**

3) Receive `INTERNET_STATUS_REQUEST_COMPLETE` event, `WaitUntilSignaledWithTerminationCheck` unblocks. Jump to next loop iteration.
4) `InternetReadFileEx` returns success with `dwBufferLength` set to zero.
5) Finally, we break out of the loop thinking the download has completed successfully.

nickdademo:
The following is my proposal to handle the `INTERNET_STATUS_CONNECTION_CLOSED` event during a download. This will make InternetReadFileEx() return failure instead of success in the above-described Step 4 (as it should).

1) Add public function `Close` to `InetHandle` struct:

```
struct InetHandle
{
    InetHandle(HINTERNET handle) : m_handle(handle), m_callback(NULL) {}

    ~InetHandle()
    {
        Close();
    }

    void SetStatusCallback(INTERNET_STATUS_CALLBACK callback)
    {
        m_callback = callback;
        InternetSetStatusCallback(m_handle, m_callback);
    }

    void Close()
    {
        if (m_handle)
        {
            if (m_callback)
                InternetSetStatusCallback(m_handle, NULL);
            InternetCloseHandle(m_handle);
        }
    }

    operator HINTERNET() const { return m_handle; }

    HINTERNET m_handle;
    INTERNET_STATUS_CALLBACK m_callback;
};
```

2) Add new case in `DownloadInternetStatusCallback`:
```
        case INTERNET_STATUS_CONNECTION_CLOSED:
            context->conn->Close();
```

It seems to be working fine - I'll do some more testing and then update this PR.

Also, I'm curious as to what behaviour you get when you kill networking connectivity during a transfer?

GitCop:

 * Commit: 1b886b899f345dd535d128fd73571570abcae563

---

vslavik:
> Also, I'm curious as to what behaviour you get when you kill networking connectivity during a transfer?

I was getting errors as expected. I think this is due to timing more than anything else and things like using Parallels, its network conditioner or removing the virtual net iface instead of unplugging may plan a role.

Anyway: thanks a lot for digging deeper into this and fixing the bug, the new version is more robust and “proper” I think (if nothing else, it works in absence of Content-Length).

Merging now, thanks again!

PR #116, opened 2016-09-20
Add background checking for updates

jaredburkeen:
Our application is a long-running application, it's intended to be run for the life of the system.
We desired to have the update-checking occur more frequently, currently the check for updates only occurs at application startup.
This change supports checking for updates periodically while the application is running.

### DIFF
--- a/src/dll_api.cpp
+++ b/src/dll_api.cpp
@@ -69,47 +69,12 @@ WIN_SPARKLE_API void __cdecl win_sparkle_init()
         // first things first
         UpdateDownloader::CleanLeftovers();
 
-        // check for updates
-        bool checkUpdates;
-        if ( Settings::ReadConfigValue("CheckForUpdates", checkUpdates) )
-        {
-            if ( checkUpdates )
-            {
-                static const time_t ONE_DAY = 60*60*24;
-
-                time_t lastCheck = 0;
-                Settings::ReadConfigValue("LastCheckTime", lastCheck);
-                const time_t currentTime = time(NULL);
-
-                // Only check for updates in reasonable intervals:
-                const int interval = win_sparkle_get_update_check_interval();
-                if ( currentTime - lastCheck >= interval )
-                {
-                    // Run the check in background. Only show UI if updates
-                    // are available.
-                    UpdateChecker *check = new UpdateChecker();
-                    check->Start();
-                }
-            }
-        }
-        else // not yet configured
-        {
-            bool didRunOnce;
-            Settings::ReadConfigValue("DidRunOnce", didRunOnce, false);
-            if ( !didRunOnce )
-            {
-                // Do nothing on the first execution of the app, for better
-                // first-time impression.
-                Settings::WriteConfigValue("DidRunOnce", true);
-            }
-            else
-            {
-                // Only when the app is launched for the second time, ask the
-                // user for their permission to check for updates.
-                UI::AskForPermission();
-            }
-        }
+        // Run the check in background. Only show UI if updates
+        // are available.
+        UpdateChecker *check = new UpdateChecker();
+        check->Start();
     }
+    
     CATCH_ALL_EXCEPTIONS
 }
 
--- a/src/updatechecker.cpp
+++ b/src/updatechecker.cpp
@@ -36,6 +36,7 @@
 #include <vector>
 #include <cstdlib>
 #include <algorithm>
+#include <winsparkle.h>
 
 using namespace std;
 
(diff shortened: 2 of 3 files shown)


GitCop:

 * Commit: 8101ba9e8586c40137eb07539c8175e0fe4bb1cb

---

vslavik:
Thanks! This would be *very* welcome addition and definitely should be done.

Unfortunately, this PR is basically unreviewable because it changes huge chunks of the code with spurious whitespace formatting changes, with real changes hidden somewhere in it. Please respect the project’s formatting (4 spaces, not tabs), i.e. use it in newly added lines *and* don’t change existing code’s formatting.

I’m not quite sure what the changes do because of it, but I can see that it shows the update info unconditionally, does it not? It IMHO should behave the way macOS Sparkle does, i.e. check for the updates and/or show the update notification only if the app is active. I.e. the user leaves it running for a long time in the background, then goes back to the app’s window and *then* the update pops up.

Also, unless I’m missing something (entirely possible), the background checking thread is never cleanly terminated, right? It arguably should be...

src/updatechecker.h
```diff
@@ -70,7 +70,8 @@ class UpdateChecker : public Thread
     virtual bool ShouldAutomaticallyInstall() const { return false; }
 
 protected:
-    virtual void Run();
```
vslavik: The use of `virtual` here is intentional, to clearly mark it as overriden (modern C++ would use `override`, I don’t use it for compatibility yet).

src/updatechecker.cpp
```diff
@@ -244,8 +242,7 @@ void UpdateChecker::Run()
 				}
 			}
 		}
-		// Check every 5 minutes
-		Sleep(300000);
+		Sleep(win_sparkle_get_update_check_interval());
```
vslavik: This is unfortunate, because it prevents the thread from being signalled in any way, e.g. to force manual check (which could then use the same thread) or to terminate. `m_terminateEvent.WaitUntilSignaled(timeout)` should be used instead.

src/updatechecker.h
jaredburkeen: Sounds good, I'll bring the virtual back.

src/updatechecker.cpp
jaredburkeen: Good call, that makes sense.

jaredburkeen:
Sorry  about the whitespace changes, that was an oversight on my part.

For our application, our customers don't really interact with a UI. There's some basic UI for config/setup, but mostly they just expect it to run in the background. So for our purposes, we need for it run regardless if the app is active. Maybe I could add a configuration option to make this behavior optional?

And I believe you're correct, I didn't add anything to cleanly kill the background update-checking thread. I'll see what I can do there.

jaredburkeen:
Ugh,looks like I introduced some whitespace problems, I'll take care of that...

vslavik:
Can you please use `git rebase` and `git push -f` to make it a clean commit, instead of a bunch of fixups like “Whitespace fixes”?

jaredburkeen:
No problem, it's cleaned up!

src/dll_api.cpp
```diff
@@ -69,47 +69,12 @@ WIN_SPARKLE_API void __cdecl win_sparkle_init()
...
-                    // Run the check in background. Only show UI if updates
-                    // are available.
-                    UpdateChecker *check = new UpdateChecker();
-                    check->Start();
-                }
-            }
-        }
-        else // not yet configured
-        {
-            bool didRunOnce;
```
vslavik: You lost all of this code somewhere.

src/updatechecker.cpp
```diff
@@ -219,9 +220,34 @@ UpdateChecker::UpdateChecker(): Thread("WinSparkle updates check")
 
 void UpdateChecker::Run()
 {
-    // no initialization to do, so signal readiness immediately
-    SignalReady();
+    while (1)
+    {
+        // no initialization to do, so signal readiness immediately
+        SignalReady();
```
vslavik: This should only be called once (and the purpose, which implies this, is documented).

src/updatechecker.cpp
```diff
@@ -219,9 +220,34 @@ UpdateChecker::UpdateChecker(): Thread("WinSparkle updates check")
...
+
+        bool checkUpdates;
+        if (Settings::ReadConfigValue("CheckForUpdates", checkUpdates))
```
vslavik: In this place, where you lost the `else` branch (which wouldn’t make sense here), it doesn’t make sense to check if the value could be read with the double-if and `Settings::ReadConfigValue("CheckForUpdates”, checkUpdates, false);`, i.e. using a default value, and ignoring the return value makes more sense.

src/updatechecker.cpp
```diff
@@ -219,9 +220,34 @@ UpdateChecker::UpdateChecker(): Thread("WinSparkle updates check")
...
+                {
+                    UpdateCheckWorker();
+                }
+            }
+        }
+        m_terminateEvent.WaitUntilSignaled(win_sparkle_get_update_check_interval());
+    }
+}
 
+void UpdateChecker::UpdateCheckWorker()
```
vslavik: This function isn’t really a “worker” (i.e. a thread that does some work), but a helper function to perform the task. `PerformUpdateCheck()` perhaps?

src/dll_api.cpp
```diff
@@ -69,47 +69,12 @@ WIN_SPARKLE_API void __cdecl win_sparkle_init()
...
-            else
-            {
-                // Only when the app is launched for the second time, ask the
-                // user for their permission to check for updates.
-                UI::AskForPermission();
-            }
-        }
+        // Run the check in background. Only show UI if updates
+        // are available.
+        UpdateChecker *check = new UpdateChecker();
```
vslavik: Hmm, this launches the updates checking thread even if the user explicitly said they don’t want that. I see the logic behind it (if the user later changes their mind, the checker thread will pick the change up), but it’s still a bit wasteful, isn’t it? 

IOW, which is the lesser evil: always needlessly creating a thread that will never be used, or re-enabling automatic checks only after the user restarts the application? I’m inclined towards the latter, but maybe I’m missing something?

vslavik:
Also, regarding this:

> For our application, our customers don't really interact with a UI. There's some basic UI for config/setup, but mostly they just expect it to run in the background. So for our purposes, we need for it run regardless if the app is active. Maybe I could add a configuration option to make this behavior optional?

I think we’ll have to do something like that. Rather than a setting, I think a callback is more appropriate way of checking for potentially more complicated conditions. There’s already `win_sparkle_set_can_shutdown_callback()` for deciding whether it’s safe to terminate the application and apply an update, maybe an analogous `win_sparkle_set_can_show_ui_callback()`? Your implementation would just return `true` and I’ll add one (to be the default) to check if the process has any visible windows.

PR #119, opened 2016-10-19
Allow static linking by using the current hInstance of the process

hpvb:
When we are statically linked we still need a hInstance from somewhere,
since DllMain() never gets called we have to rely on a the one of the
current module.

This works at least for our use in ScummVM.

### DIFF
--- a/src/ui.cpp
+++ b/src/ui.cpp
@@ -1457,7 +1457,11 @@ void UI::Run()
     // HINSTANCE of this DLL, not of the main .exe.
 
     if ( !ms_hInstance )
-        return; // DllMain() not called? -- FIXME: throw
+    {
+        // If DllMain() was not called, assume we're statically linked
+        // and use the hInstance of the containing program.
+        ms_hInstance = GetModuleHandle(NULL);
+    }
 
     // IMPLEMENT_WXWIN_MAIN does this as the first thing
     wxDISABLE_DEBUG_SUPPORT();


GitCop:

 * Commit: ccb6dc657f6d6834253f974dd54ec8130d68cbd3

---

src/ui.cpp
```diff
@@ -1456,8 +1456,11 @@ void UI::Run()
     // We need to pass correct HINSTANCE to wxEntry() and the right value is
     // HINSTANCE of this DLL, not of the main .exe.
 
-    if ( !ms_hInstance )
-        return; // DllMain() not called? -- FIXME: throw
+    if ( !ms_hInstance ) {
```
vslavik: Can you please fix this to respect upstream code formatting, i.e. `{` on its own line (see above your changes)? TIA!

vslavik:
I’d advise against static linking (may break in the future), but I’m sure you know what you’re doing — and there’s definitely no harm in applying this.

hpvb:
I have fixed the whitespace issues. I kept the braces, even though the function is only one line because I find comments in a single line abbreviated block to be very sensitive to errors.

If you disagree I will amend again removing the braces.

vslavik:
> If you disagree I will amend again removing the braces.

No, you’re right about that (and even if I disagreed, that would be far less jarring).

Thanks for the PR and sorry for the delay, it’s been a busy week :(

PR #120, opened 2016-10-19
Allow building against WX3.1.0+

hpvb:
The function wxMSWDisableSettingHighDPIAware() no longer exists in
WxWidgets 3.1.0 and later. Futhermore it only existed from 3.0.3
onwards.

This PR adds some compile time checks to disable the call entirely when
an inappropriate version of wx is detected.

From the 3.0.3 code:
since 3.0.3, but only available in 3.0.x, not 3.1+ which doesn't make
the SetDPIProcessAware() call anymore.

### DIFF
--- a/src/ui.cpp
+++ b/src/ui.cpp
@@ -1473,7 +1473,9 @@ void UI::Run()
     SignalReady();
 
     // Run the app:
+#if wxCHECK_VERSION(3, 0, 3) && !wxCHECK_VERSION(3, 1, 0)
     wxMSWDisableSettingHighDPIAware();
+#endif
     wxEntry(ms_hInstance);
 }
 


vslavik:
I was conflicted on this one.

On one hand, why not, if it makes your life easier. (Personally, I don’t bother and ship normal config of WinSparkle with my wx-based application; it’s compressed size is minuscule compared to the rest.)

On the other hand, though:

1. There are reasons why WinSparkle bundles wx, and avoiding messy `#ifdefs` is one of them.
2. It would be preferable to simply (well…) upgrade WinSparkle’s dependency to 3.1 (which is unstable in API sense, but fine).
3. I’m uncomfortable with the whole custom compilation with unknown wx version thing. The plan is to support ZIP updates too, not just re-running the installer, and those require WinSparkle to move itself out of the way… which it can’t done if it has extra dependencies.

In the end, I’m going to merge this one, but I can’t promise that the code will work with out-of-tree wx forever, and it’s certainly unsupported.

hpvb:
We also statically link to wx for sparkle, however we don't want to ship any code we did not build ourselves. We could of course just use an older version of wx for everything but I'm also not so sure that's the best idea (we ship wx as part of scummvm-tools also).

To simplify our build process only having to deal with one wx is preferable.

vslavik:
> We also statically link to wx for sparkle, however we don't want to ship any code we did not build ourselves.

I’m the same way, compiling everything, never using vendor binaries. But as a rationale for this PR, this makes no sense — you build WinSparkle yourself, don’t you? WinSparkle’s bundled wx is *also built from sources* as part of WinSparkle's build too…

Anyway, I momentarily forgot that you’re also custom-linking WinSparkle statically, in which case yes, you need this — and not only is dealing with a single wx instance preferable, it’s all but *necessary* if you link everything statically together. I’ll need to keep this use-case in mind...

hpvb:
We cross-compile on Linux and can't use the project file, we use a custom makefile and we link against the same version of WX as we use to ship scummvm-tools.

Our setup isn't entirely standard :)

PR #126, opened 2017-01-20
Wait in background and trigger update on interval

kalaspuffar:
I read the pull request #116 and wanted to help out doing the additional changes required in order to make it a part of the library.

We implemented a background process program that does work for the user on both Mac and PC. And  we just configured the application for auto updating on both platforms and were baffled when the update never happened.

Today we implemented this by using QTimer to update the application every hour by calling win_sparkle_check_update_without_ui().

But a better solution would be to have this function as a part of the library.

I copied a diff from the original pull request because the original authors repository wasn't open so I couldn't fork it and then I made the changes you requested. I think all of them are reasonable and I don't really get why code was removed but your original code is a good addition so I just restored it and changed the calls.

Please tell me if you require any other changes or if you want to discuss the issue.

### DIFF
--- a/src/dll_api.cpp
+++ b/src/dll_api.cpp
@@ -75,21 +75,8 @@ WIN_SPARKLE_API void __cdecl win_sparkle_init()
         {
             if ( checkUpdates )
             {
-                static const time_t ONE_DAY = 60*60*24;
-
-                time_t lastCheck = 0;
-                Settings::ReadConfigValue("LastCheckTime", lastCheck);
-                const time_t currentTime = time(NULL);
-
-                // Only check for updates in reasonable intervals:
-                const int interval = win_sparkle_get_update_check_interval();
-                if ( currentTime - lastCheck >= interval )
-                {
-                    // Run the check in background. Only show UI if updates
-                    // are available.
-                    UpdateChecker *check = new UpdateChecker();
-                    check->Start();
-                }
+                UpdateChecker *check = new UpdateChecker();
+                check->Start();
             }
         }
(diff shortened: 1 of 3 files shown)


vslavik:
Thanks for picking this up! I’ll try to have a closer look over the weekend; things are a bit hectic around here these days, so feel free to bug me if I don’t.

kalaspuffar:
@vslavik Been sick over the weekend so haven't bothered you but when you have time please look into this. I guess there might be more people than me that think this function is essential to the winsparkle. I don't know if the function of only doing updates when active as suggested but that might be an addition in the future. To have the option of regularly check for updates should improve the library.

I've implemented this feature with a workaround as I said so I don't have any real hurry to get this pulled but others might not be as fortunate.

vslavik:
Shit, I forgot to push from my working copy. Upon review, your commit was perfect, I only added some tuning of the wait behavior in f0ddddc274cf94fcfaa0bd4ee7af094fc12e4515 (which I started in January, forgot all about and finished now).

Let my know if you have any remaining issues with this when you update WinSparkle. Out of curiosity, is your UI mostly hidden as in #116 too or does the check to see if the app is active (i.e. don't prompt with updates unless the user is using the app) make sense in your case?

Thanks!

kalaspuffar:
@vslavik Well our application was checking for updates in the background and only prompted the user if there was an update.

Sadly the project we used winsparkle has been retired as we found a better solution for the users. But it's good to know that we can pick up winsparkle for the next project we require an update process to.

Best regards
Daniel

PR #133, opened 2017-03-03
Remove Query String From URL Filename

### LINKED ISSUE #131: Error getting temp filename
I am using WinSparkle 0.5.3 and when I try to download an update I get an error. I think the error is because of an invalid filename for the temp file name (similar to  #109) I am using WinSparkle in a C# app (probably irrelevant) and am downloading the updates via a redirect (similar to hockeyapp). My url is something like files.myorg.com/myfile.msi which then is redirected to an Azure Blob with a Shared Access Signature good for a certain amount of time. This means that the url is more like blobs.azure.net/myfile.msi?pv=fd&sdfd&sfdd&etc

The app feed rss downloads fine through this method (as does the file) but when trying to save it to the file system, Sparkle errors out (I think because the file is invalid). In the debugger, I noticed that this code returns the filename with the Shared Access Signature still on it. 

```
std::wstring GetURLFileName(const char *url)
{
    const char *lastSlash = strrchr(url, '/');
    const std::string fn(lastSlash ? lastSlash + 1 : url);
    return AnsiToWide(fn);
}
```

Then, the code errors out in this block:
```
virtual void SetFilename(const std::wstring& filename)
    {
        if ( m_file )
            throw std::runtime_error("Update file already set");

        m_path = m_dir + L"\\" + filename;
        m_file = _wfopen(m_path.c_str(), L"wb");
        if ( !m_file )
            throw std::runtime_error("Cannot save update file");
    }
```

lehne:
Strips out the query string from the URL filename as a '?' is an invalid
character in Windows filenames. This fixes #131

### DIFF
--- a/src/download.cpp
+++ b/src/download.cpp
@@ -128,7 +128,7 @@ std::wstring GetURLFileName(const char *url)
 {
     const char *lastSlash = strrchr(url, '/');
     const std::string fn(lastSlash ? lastSlash + 1 : url);
-    return AnsiToWide(fn);
+    return AnsiToWide(fn.find_first_of('?') != std::string::npos ? fn.substr(0, fn.find_first_of('?')) : fn);
 }
 
 struct DownloadCallbackContext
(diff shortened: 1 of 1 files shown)


vslavik:
Thank you!

PR #139, opened 2017-06-13
[download] - don't close handle when receiving INTERNET_STATUS_CONNECTION_CLOSED

Memphiz:
This PR fixes 302 redirects for our mirror system (http://mirrors.kodi.tv).

We are using the following appcast.xml (which you can use easily in your "example_psdk" for reproduction aswell): https://raw.githubusercontent.com/xbmc/SparkleBackend/fix_redirects_demo/sparkleupdate.xml

This announces this file for download:

http://mirrors.kodi.tv/test-builds/win32/KodiSetup-20170611-bb3345af02-sparkle-x86.exe

The mirrors are using mirrorbits and are redirecting to the actual mirror with a 302 found status. (URL redirection).

The dwInternetStatus in the callback when trying to download that file is as follows:

INTERNET_STATUS_HANDLE_CREATED
INTERNET_STATUS_DETECTING_PROXY
INTERNET_STATUS_RESOLVING_NAME
INTERNET_STATUS_NAME_RESOLVED
INTERNET_STATUS_CONNECTING_TO_SERVER
INTERNET_STATUS_CONNECTED_TO_SERVER
INTERNET_STATUS_SENDING_REQUEST
INTERNET_STATUS_REQUEST_SENT
INTERNET_STATUS_RECEIVING_RESPONSE
INTERNET_STATUS_RESPONSE_RECEIVED
INTERNET_STATUS_CLOSING_CONNECTION
INTERNET_STATUS_CONNECTION_CLOSED <- this one is the Problem because in the original callback handler it closes the client side handle
INTERNET_STATUS_REDIRECT
INTERNET_STATUS_DETECTING_PROXY
INTERNET_STATUS_RESOLVING_NAME
INTERNET_STATUS_NAME_RESOLVED
INTERNET_STATUS_CONNECTING_TO_SERVER
INTERNET_STATUS_CONNECTED_TO_SERVER
INTERNET_STATUS_SENDING_REQUEST
INTERNET_STATUS_REQUEST_SENT
INTERNET_STATUS_RECEIVING_RESPONSE
INTERNET_STATUS_RESPONSE_RECEIVED
INTERNET_STATUS_REQUEST_COMPLETE

This means we are operating on an invalid/closed handle after the request was completed and we try to download the file.

Closing the handle in the callback is not needed at all. Because of the destructor of the struct InetHandle
will close the handle as soon as the InetHandle conn goes out of scope from the DownloadFile method.

This change allows us to use winsparkle for our project with the big mirror system as backend.

Thx for considering a merge.

Regards Memphiz of Team Kodi.

### DIFF
--- a/src/download.cpp
+++ b/src/download.cpp
@@ -158,10 +158,6 @@ void CALLBACK DownloadInternetStatusCallback(_In_ HINTERNET hInternet,
         case INTERNET_STATUS_REQUEST_COMPLETE:
             context->eventRequestComplete.Signal();
             break;
-
-        case INTERNET_STATUS_CONNECTION_CLOSED:
-            context->conn->Close();
-            break;
     }
 }
 


vslavik:
> kodi.tv

As a big fan of Kodi, that's one heck of pleasure to see! Will you need a fixed release or do you use `master`?

While the current code is certainly wrong, as you demonstrated, this fix breaks *other* things by undoing part of the changes from #114 (da4bd8ccf3a9a4884e48546c54975152db755b2b). Basically we *also* need to detect unexpectedly closed connections...

vslavik:
Sorry I took so long to come up with something. Would you have a copy (or better, direct link) of the feed that still works? The above download is 404 now...

Memphiz:
https://raw.githubusercontent.com/xbmc/SparkleBackend/fix_redirects_demo/sparkleupdate.xml <- i hope this works - ensure to use the win32 / x86 path to get a valid url from the feed.

vslavik:
Thanks!

In my testing, the above b9f988a commit (now on `master`) fixes the bug. If you could check it too to make sure, that would be appreciated — and of course, if it turns out not to be enough at any time, do let me know.

Memphiz:
Will do - the change looks sane for  sure - thx for doing it properly :)

vslavik:
> the change looks sane for sure - thx for doing it properly

I don't have *that* kind opinion of it — still seems like a kludge to me and the reason I'm asking for additional testing is because I'm afraid it may still fail in case of some unfortunate timing. 

I'll release 0.5.4 with the fix as is, because it's definitely better then before, but it may not be the end of it yet...

Memphiz:
I can confirm that your fix works for us ...

vslavik:
@Memphiz Could you try if redirects still work for you with the changes from 78f86546? That one should finally be *properly* done, but I'd prefer a confirmation that I didn't break anything for you before I make a release… TIA!

Memphiz:
@vslavik this still does work. Can you release a Version with this so we can use it in kodi?

vslavik:
@Memphiz Thanks for confirmation — I just released [0.5.7](https://github.com/vslavik/winsparkle/releases/tag/v0.5.7) with the fix.

Memphiz:
thx a lot.

PR #146, opened 2017-07-31
Fix for update check not triggered

### LINKED ISSUE #144: win_sparkle_init doesn't trigger update check on first startup (regression)
**Description:**
In `0.5.3`, calling `win_sparkle_init()` results in performing a silent check. With `0.5.4` this doesn't happen anymore - no window is displayed apart from settings used, no callback functions are called.

**Reproducing:**
Master and latest release (`0.5.4`) are affected. `0.5.3` seems to work just fine. Can be reproduced with both examples provided in the repository.

Had no time to check which commit breaks it, sorry. Can do that tomorrow (maybe even provide a fix).

cloudshao:
If the LastCheckTime registry value was not set, it would default to
current time, meaning that a check wouldn't happen until after the
interval passed. If the user always closes the app before reaching
the interval, then a check would never happen.

This simply reverts back to previous behavior. Give LastCheckTime a
default value of 0 so that a check happens immediately if there's
been no previous check.

Fixes #144

### DIFF
--- a/src/updatechecker.cpp
+++ b/src/updatechecker.cpp
@@ -235,7 +235,7 @@ void UpdateChecker::Run()
         if (checkUpdates)
         {
             const time_t currentTime = time(NULL);
-            time_t lastCheck = currentTime;
+            time_t lastCheck = 0;
             Settings::ReadConfigValue("LastCheckTime", lastCheck);
 
             // Only check for updates in reasonable intervals:


GitCop:

 * Commit: 7bae3589bca30853888fac203c2556e65bad1660

---

GitCop:

 * Commit: bc102ba267d3eda8cb4b2ff3c1d12b9c567cca56

---

src/updatechecker.cpp
```diff
@@ -234,13 +234,13 @@ void UpdateChecker::Run()
 
         if (checkUpdates)
         {
-            const time_t currentTime = time(NULL);
-            time_t lastCheck = currentTime;
```
vslavik: Isn't the simplest possible fix to just do `time_t lastCheck = 0` here instead of your larger diff? Am I missing something (again)?

cloudshao: Yup, you're right. Will push a more minimal change.

cloudshao:
Pushed a more minimal change based on your comment @vslavik

vslavik:
Thanks a lot!

Issue #151, opened 2017-12-06
Update window not visible when app main window is minimized


johnmurphy01:
We are running a WPF application that currently uses WinSparkle(version `0.5.7`). We are following the recommended way to set up auto updates:

```
win_sparkle_set_app_details("Company", "Application", string.Format("{0}", System.Reflection.Assembly.GetExecutingAssembly().GetName().Version.ToString()));
win_sparkle_set_appcast_url("http://sparkleurl.com");
win_sparkle_set_update_check_interval(3600);
win_sparkle_set_automatic_check_for_updates(1);
win_sparkle_init();
```

When we launch the application, WinSparkle does indeed display the update UI. However, if we then minimize our application and then WinSparkle attempts a subsequent auto update check, the window is not available. It's as if the update UI is somewhere out of the bounds of the desktop viewport.

At the point that this issue occurs, the only way to "fix" it is to close all windows associated with our application.

We weren't sure if this was a regression or a new issue since we saw this issue(somewhat related) https://github.com/vslavik/winsparkle/issues/46

vslavik:
Please [follow these instructions](http://www.chiark.greenend.org.uk/~sgtatham/bugs.html). If you submit a bug report, presumably it's because you'd like something to be *done* about the bug. That's only possible if you describe the bug well, i.e. when you, crucially, provide *reproduction instructions*: at minimum, that includes *full code* (ideally in a form of patch against WinSparkle example) and accurate, detailed, exact steps to perform to reproduce the issue. 

That you use WPF is in all likelihood irrelevant. What exactly your secret application code does, what windows it has, with what properties etc., on the other hand, is all highly relevant and probably explanatory — but unfortunately missing from this issue.

As you observed, #46 may be relevant. Therefore the most logical first thing to do is to use a debugger  and investigate why the code that fixed #46 doesn't work — that's straightforward and not particularly time-consuming work. Because as of now you are the only person with access to reproduction case, you are also the only person that can actually *do* it and provide some insight. Can you please do it and share any findings (ideally, of course, in the form of a patch)?

Thanks!

johnmurphy01:
Hopefully this will be enough information for you to replicate and identify the bug. I am providing a zip file containing a sample application that makes the same WinSparkle calls that our production application makes. Here are the replication steps:

- Launch the sample application
Observe that the WinSparkle update UI will appear with the new version.

- Click "Remind me Later"
[...]

PR #152, opened 2017-12-06
Ensure that biggest window is on a valid display

johnmurphy01:
Referencing issue #151:

When the application window is minimized, the resultant "biggest" window ends up being positioned off of the view-able display(somewhere around `x` of -32000 and `y` of -32000). Our solution is to make sure that the biggest window within a valid display boundary before setting `data.biggest`. If the check fails, then `data.biggest` becomes empty and the updater UI is subsequently positioned in the center of the main display.

### DIFF
--- a/src/ui.cpp
+++ b/src/ui.cpp
@@ -183,7 +183,7 @@ BOOL CALLBACK EnumProcessWindowsCallback(HWND handle, LPARAM lParam)
     RECT rwin;
     GetWindowRect(handle, &rwin);
     wxRect r(rwin.left, rwin.top, rwin.right - rwin.left, rwin.bottom - rwin.top);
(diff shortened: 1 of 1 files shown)


vslavik:
I replaced this with a cleaner check for the specific problem (minimized window, -32000 as coordinates) in e6f62a1113a9b90df576f6245e232c7219b34343, so for your specific situation, this PR is no longer necessary. I think it should *also* be applied, though, because it is more robust in the sense that it gracefully handles it if some other unanticipated situation happens.

vslavik:
Committed the above display check as f44b174a06c453e6c80eb99a2637ec4fad203f16 (under your name, because I only cosmetically rearranged).

PR #156, opened 2017-12-11
Free web browser after display

johnmurphy01:
Free the web browser object after showing the update window. This allows the initialization code for the web browser to run again if a subsequent check for updates process is executed.

**Added Note:** This issue does appear to _only_ affect release notes that are loaded using the `description` tag. Loading release notes using a url does not appear to have this issue.

### DIFF
--- a/src/ui.cpp
+++ b/src/ui.cpp
@@ -878,6 +878,8 @@ void UpdateDialog::StateUpdateAvailable(const Appcast& info, bool installAutomat
     // take some time to load the MSIE control:
     if ( showRelnotes )
         ShowReleaseNotes(info);
+
+	m_webBrowser->Free();
 }
 
 
--- a/src/ui.cpp
+++ b/src/ui.cpp
@@ -879,7 +879,7 @@ void UpdateDialog::StateUpdateAvailable(const Appcast& info, bool installAutomat
     if ( showRelnotes )
         ShowReleaseNotes(info);
 
-	m_webBrowser->Free();
+	m_webBrowser.Free();
 }
 
 


src/ui.cpp
```diff
@@ -878,6 +878,8 @@ void UpdateDialog::StateUpdateAvailable(const Appcast& info, bool installAutomat
     // take some time to load the MSIE control:
     if ( showRelnotes )
         ShowReleaseNotes(info);
+
```
vslavik: Nitpicking: This appears to use tabs instead of 4 spaces (and would seem to add trailing whitespace on this line too).

johnmurphy01: Hah, been caught by this before. I'll fix

vslavik:
This fix seems wrong — it doesn't really address the (still unclear) underlying problem. I suspect the actual bug is somewhere in the section that starts with `// If the code below looks crazy` so it would be preferable to fix there.

But even if destroying and recreating the control is the only fix (there's only so much tolerance for debugging`IWebBrowser`, I get that), it's incorrect in that it doesn't destroy the accompanying `wxActiveXContainer` instance. So you're effectively having two of them in place now, one hiding another (and the browser control not actually destroyed, but unused)...

vslavik:
Does this alternative patch work for you (it does in my testing)?

```diff
diff --git a/src/ui.cpp b/src/ui.cpp
index ef0ce53..df2f5d2 100644
--- a/src/ui.cpp
+++ b/src/ui.cpp
@@ -999,6 +999,15 @@ void UpdateDialog::ShowReleaseNotes(const Appcast& info)
         m_webBrowser = browser;
 
         new wxActiveXContainer(m_browserParent, IID_IWebBrowser2, browser);
+
+        m_webBrowser->Navigate
+        (
+            wxBasicString("about:blank"),
+            NULL,  // Flags
+            NULL,  // TargetFrameName
+            NULL,  // PostData
+            NULL   // Headers
+        );
     }
 
     if( !info.ReleaseNotesURL.empty() )
@@ -1014,15 +1023,6 @@ void UpdateDialog::ShowReleaseNotes(const Appcast& info)
     }
     else if ( !info.Description.empty() )
     {
-        m_webBrowser->Navigate
-                      (
-                          wxBasicString("about:blank"),
-                          NULL,  // Flags
-                          NULL,  // TargetFrameName
-                          NULL,  // PostData
-                          NULL   // Headers
-                      );
-
         HRESULT hr = E_FAIL;
         IHTMLDocument2 *doc;
         hr = m_webBrowser->get_Document((IDispatch **)&doc);
```

If not, could you please share the feed's content or URL?

johnmurphy01:
I'll run my tests with that fix and let you know

johnmurphy01:
That appears to resolve the issue! Would you like me to add this to my PR or are you going to make the changes directly to your `master` branch?

vslavik:
No need, I already had it in my working tree, so I committed it now. Thanks!

Issue #23, opened 2014-05-27
Add support for DSA signed updates


vslavik:
You mention in your docs that signing updates is not yet supported, but I didn't see this in the planned features list. So hereby I am requesting this feature.

Background: We are considering using Sparkle for ScummVM ( http://www.scummvm.org) on Mac OS X. It then would make a lot of sense to also use WinSparkle for our Windows port. But I would feel very uncomfortable about installing unsigned updates automatically this way. (Granted, we don't yet provide signatures for our regular downloads either, but this, too, is something that should change).

Hopefully this won't be too difficult to implement. Microsoft seems to provide an API for verifying crypto signatures (including DSA support), according to MSDN ( http://msdn.microsoft.com/en-us/library/aa387761%28v=VS.85%29.aspx)

treasurebox:
Slightly related question: Does WinSparkle support downloading the xml from https:// _but_ only if ssl cert is okay? 
I couldn't find any mention about this about the doc, but either executable signature checking or https xml fetching seems to be a minimum for not opening a backdoor on user's machine.
Thank You for the quick answer, and the awesome work so far!

vslavik:
> related

That word doesn’t mean what you think it means. Please don’t post completely off-topic comments like this.

> Does WinSparkle support downloading the xml from https:// but only if ssl cert is okay? 

You have the code at your disposal, why don’t you verify it yourself and if there’s any issue, submit a patch? 

> for the quick answer

This is a volunteer, community project, please don’t treat it as a commercial offering with support. Instead, *contribute* — even if it’s “only” information or documentation improvements.

treasurebox:
I think it's not off topic at all. The same people who think they need DSA signed updates might just be satisfied with https based xml download.
Anyway, looking into the code, it uses InternetOpenUrl(), which is not very well documented in this regard, but trying it out (see http://stackoverflow.com/questions/29545544/is-internetopenurl-function-on-windows-secure-enough-if-not-how-to-make-it-str/29576201#29576201), it seems that it indeed checks https validity by default. So the answer to my own question is yes.

vslavik:
> I think it's not off topic at all.

This issue is for tracking support for DSA signatures. You ask a question about validation of certificates, which is something that has *zero* overlap with the issue’s subject, technically or thematically. You are off topic *by definition*. By your the-same-people logic, *anything* about WinSparkle would be on-topic in *any* issue...

mm-akruse:
[...]

PR #157, opened 2017-12-17
Add DSA signature verification (OpenSSL based)

Youw:
Compatible with Sparkle method of DSA signing and verification.

Uses OpenSSL and adds external dependency: `libeay32.dll` (but NOT `ssleay32.dll`).

Theoretically can be reimplemented, to use WinAPI only.

Reffers to #23.


GitCop:

 * Commit: 458695052883e3bce413c25ff5e17b093261e924

 * Commit: 6119e2e2df6395a6790b5759d8f68339d7b1f388

---

vslavik:
Thank you *a lot!*

Just two quick question before I actually look at the code:

> Compatible with Sparkle method of DSA signing and verification.

To confirm, this means that a signature produced by Sparkle's macOS tools will be accepted, right?

> Uses OpenSSL and adds external dependency: libeay32.dll (but NOT ssleay32.dll).

External *DLL* is unfortunate (I misunderstood this bit when we talked about adding an external dependency), but I think I have a fix for that: I use a VC++ project for statically-built libcrypto in something else, so I'll follow this PR with changes to use that and return into single-DLL packaging. Might be worth considering BoringSSL ([vcxproj](https://chromium.googlesource.com/external/github.com/grpc/grpc/+/chromium-deps/2016-07-19/vsprojects/vcxproj/boringssl)) or LibreSSL for that too, but again: don't worry about that part.

Did you use 1.0 or 1.1 series of OpenSSL?

Youw:
> this means that a signature produced by Sparkle's macOS tools will be accepted

Right. And vice versa.

> Did you use 1.0 or 1.1 series of OpenSSL?

Good catch. I probably should have mention that in README.
I've used OpenSSL 1.0.2.

src/signatureverifier.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+#include "signatureverifier.h"
+
+#include <stdexcept>
+
+#define WIN32_LEAN_AND_MEAN
+#include <Windows.h>
+#include <Wincrypt.h>
+
+#pragma comment(lib, "crypt32.lib")
```
Youw: Missed this - it is for Base64ToBin.
Does WinSparkle support MinGW/Cygwin builds?

Youw:
BTW: let me know, if you require any additional assistance from me.

Youw:
I've managed to port signature verification to bare WinAPI (so no `libeay32.dll`), but there is another problem: [CryptoAPI does not support DSA keys larger that 1024](https://msdn.microsoft.com/en-us/library/windows/desktop/bb931357(v=vs.85).aspx).

I have perfectly working code in [my repo](https://github.com/Youw/winsparkle/tree/win_crypto_dsa_verify), but it can use only DSA keys 1024 bits large. It partially breaks compatibility with Sparkle (and potentially less secure).

There is also a new API from Microsoft - [CNG](https://msdn.microsoft.com/en-us/library/windows/desktop/aa376210(v=vs.85).aspx), but it is available only starting from Windows Vista and DSA keys larger than 1024 supports only [starting from Window 8](https://msdn.microsoft.com/en-us/library/windows/desktop/aa375534(v=vs.85).aspx).

So, the only way to be fully compatible with Sparkle (and FIPS compliant) is to use OpenSSL on Windows.

Alternatives:
1) Use 1024 bit key - [security is no longer guaranteed with 1024 bit long keys](https://stackoverflow.com/a/16935946/3697939);
2) Switch to 4096 RSA signatures - secure, but fully breaks compatibility with Sparkle project.

@vslavik, thoughts?

src/signatureverifier.cpp
vslavik: Somewhat — there's the cmake build system that is meant to work with it, but it's not maintained by me. In any case, that's a trivial thing to `#ifdef`, so don't worry about that.

vslavik:
First of all, sorry for not reviewing this before Xmas as I hoped to — and thanks a lot for your patience and continued work. I'll do my best to merge this this week.

Re DSA: wow, I had no idea of this mess.

So yes, this is an argument for sticking to OpenSSL. Another alternative is

3. Have both implementations and document that if you support anything other than Windows 8, you have to stick to 1024 keys — an acceptable tradeoff for supporting old OSes. If Windows 7 was the cutoff, I think I would be OK with that, but Windows 8 is still a bit high requirement. 

Finally, Sparkle has an open ticket about migrating off DSA at https://github.com/sparkle-project/Sparkle/issues/1037. That opens another option:

4. Agree with Sparkle guys on what to use and how and implement only that, i.e. being compatible with latest/future Sparkle.

On the other hand: you have already done all the hard work to support DSA via OpenSSL. Would be shame to waste that. On the *third* hand, the moment we release DSA support, we're stuck with supporting it forever.

src/signatureverifier.cpp
Youw: I just don't feel right, if I keep it as is. [9d811dd](https://github.com/Youw/winsparkle/commit/9d811dd2840ce6795d17ca870514a01d523834a2) =)

Youw: BTW: cmake build is broken after one of the expat updates

vslavik: @Youw Pro-tip: when you find yourself writing "BTW" in an issue tracker, stop right there — you're being off-topic. If you have a *completely unrelated* issue, file an unrelated issue in the tracker. In the case of cmake build, make that a PR (see the [explanation](https://github.com/vslavik/winsparkle/tree/master/cmake) why).

vslavik: @Youw I'm sorry, let me rephrase: since the cmake build is unsupported, you don't have to worry about it as part of this PR.

Youw: Sure, totally agree.

src/settings.h
```diff
@@ -111,6 +111,15 @@ class Settings
         return ms_registryPath;
     }
 
+    /// Return DSA public key to verify update file signature
+    static std::string GetDSAPubKeyPem()
+    {
+        CriticalSectionLocker lock(ms_csVars);
+        if (ms_DSAPubKey.empty())
+            SetDSAPubKeyPem(GetCustomResource("DSAPub", "DSAPEM"));
```
vslavik: I know (now…took me writing first version of this comment to realize it) that this works, because Windows critical sections are re-entrant, but I think it's not very intuitive — I expect this to deadlock because `SetDSAPubKeyPem` also holds `ms_csVars`.

I would prefer to do this in the same way as `GetAppcastURL()` above, i.e. just assign into `ms_DSAPubKey`. This  would have consequences for other code, of course — I'll comment on them in their respective places.

src/signatureverifier.cpp
```diff
@@ -0,0 +1,399 @@
...
+        }
+
+    }; // DSAWrap
+
+    DSAPub m_dsaPub;
+    DSAPub &PubKey()
+    {
+        if (static_cast<DSA*>(m_dsaPub) == NULL)
+        {
+            std::string pem = Settings::GetDSAPubKeyPem(); // side effect - it will call SignatureVerifier::SetDSAPubKeyPem
```
vslavik: Such side-effects should be, generally speaking, avoided.

src/updatedownloader.cpp
```diff
@@ -177,6 +179,14 @@ void UpdateDownloader::Run()
       UpdateDownloadSink sink(*this, tmpdir);
       DownloadFile(m_appcast.DownloadURL, &sink, this);
       sink.Close();
+
+      if (!SignatureVerifier::DSASHA1SignatureValid(sink.GetFilePath(), m_appcast.DsaSignature))
+      {
+          // remove potentially corrupted file
+          _wremove(sink.GetFilePath().c_str());
```
vslavik: Isn't it better to just call `CleanLeftovers()` (which does full cleanup) here?

src/updatedownloader.cpp
```diff
@@ -177,6 +179,14 @@ void UpdateDownloader::Run()
       UpdateDownloadSink sink(*this, tmpdir);
       DownloadFile(m_appcast.DownloadURL, &sink, this);
       sink.Close();
+
+      if (!SignatureVerifier::DSASHA1SignatureValid(sink.GetFilePath(), m_appcast.DsaSignature))
```
vslavik: This block should only be called if a public key is set. This is the promised comment on how I think setting the public key should be done better:

1. Only store it in `Settings`, don't do anything extra.
2. Here, obtain the key from `Settings:: GetDSAPubKeyPem()`. If not empty (but regardless of whether `m_appcast.DsaSignature` is empty or not, because that one is untrusted), call this block, otherwise don't. 
3. Pass the public key as argument to `DSASHA1SignatureValid`, or set it here, rather than doing that at some earlier point.

That would have the advantage that it would also work without OpenSSL DLLs if signatures are not used (which as far as I can tell is not the case with the current code due to setting the key possibly earlier?).

src/signatureverifier.cpp
```diff
@@ -0,0 +1,399 @@
...
+
+        DSAPub &pubKey(PubKey());
+
+        const int code = DSA_verify(0, sha1, ARRAYSIZE(sha1), (const unsigned char*)signature.c_str(), signature.size(), pubKey);
+
+        if (code == -1) // OpenSSL error
+            throw std::runtime_error(ERR_error_string(ERR_get_error(), nullptr));
+
+        if (code != 1)
+            throw std::runtime_error("DSA Signature not match!");
```
vslavik: Let's use just "Invalid signature".

src/updatedownloader.cpp
Youw: Sure - haven't seen it.

src/updatedownloader.cpp
Youw: OK, I got the idea (basically we keep backward compatibility)
> don't do anything extra

Maybe I'll remove key caching, but replace it with key verification
It will also address `side effect` issue

Youw:
Thanks for comments. I'll update this PR within couple of days.
So I guess you're not going to add static libcrypto, and leave it as is?
> Also, could you send me (my email is in my github info) the private key you used so that I can properly sign the example download at winsparkle.org?

I thought you might want to generate your own keypair, and replace this public key instead.
But I'll send it to you anyway.

vslavik:
> So I guess you're not going to add static libcrypto, and leave it as is?

I'll make the change to use static lib later, so that I don't interfere with your changes. 

> I thought you might want to generate your own keypair, and replace this public key instead.

I wanted to avoid unneeded commits...

GitCop:

 * Commit: 6cbf777f44c093ed3621ca9be4861c5518c7efae

---

vslavik:
I pushed static OpenSSL build to the branch.  The size difference of `WinSparkle.dll` is not as bad as I feared it would be:

* version from master is 1,387,520 bytes
* signature-verifying statically compiled version is 1,894,400 bytes

That's a ~36%, ~500 kB difference, I consider that acceptable. Ideally I'd investigate whether it could be made smaller with [libtomcrypt](https://github.com/libtom/libtomcrypt) (being my preference due to reputation, small footprint and public domain license), [Botan](https://botan.randombit.net) or [Crypto++](https://github.com/weidai11/cryptopp) to cut on dependency size, but 500 kB is fine for 0.6.0.

src/signatureverifier.cpp
```diff
@@ -259,14 +203,6 @@ class TinySSL
 private:
     static HMODULE ms_libeay32;
```
Youw: remove it too

src/signatureverifier.cpp
```diff
@@ -329,14 +265,6 @@ class TinySSL
 
 HMODULE TinySSL::ms_libeay32 = NULL;
```
Youw: remove

GitCop:

 * Commit: ea0a09554885d2b64ce41c4abaa6689bcdb92b13

---

src/signatureverifier.cpp
vslavik: Thanks!

GitCop:

 * Commit: ea0a09554885d2b64ce41c4abaa6689bcdb92b13

 * Commit: 92f526a4378ff8f1f13d4675c7ded5965830883b

---

vslavik:
I added one more thing: dedicated message for signature failure, as Sparkle does. It is important enough failure state to specifically alert the user to. I think we're good to go with this.

Thanks again for your work on this PR, it's much appreciated!

Youw:
Nicely done!

I think it also makes sense to use `BadSignatureException` [here](https://github.com/vslavik/winsparkle/pull/157/commits/92f526a4378ff8f1f13d4675c7ded5965830883b#diff-58370b0e0e98f68e86ee09546e10b2eeL216), [here](https://github.com/vslavik/winsparkle/pull/157/commits/92f526a4378ff8f1f13d4675c7ded5965830883b#diff-58370b0e0e98f68e86ee09546e10b2eeL245) and [here](https://github.com/vslavik/winsparkle/pull/157/commits/92f526a4378ff8f1f13d4675c7ded5965830883b#diff-58370b0e0e98f68e86ee09546e10b2eeL281).

I guess everywhere in `signatureverifier.cpp`.

vslavik:
> I think it also makes sense to use BadSignatureException here, here and here.

My reasoning for using it only in a few places was that it should be used when the code is (reasonably) certain the issue is a bad signature. If the problem is something else, such as invalid input, existing exception classes are more suitable (e.g. `invalid_argument`).

Youw:
Good point.

I missed the fact that [this](https://github.com/vslavik/winsparkle/pull/157/commits/92f526a4378ff8f1f13d4675c7ded5965830883b#diff-58370b0e0e98f68e86ee09546e10b2eeR309) fixes all issues I was wanted to avoid.

But there is still one more issue worth considering: logging.
Maybe it worth:
```
    catch (BadSignatureException&)
    {
        throw;
    }
    catch (const std::exception &e)
    {
        throw BadSignatureException(e.what());
    }
    catch (...)
    {
        throw BadSignatureException("Unknown signature verification error");
    }
```

Right now, it `dsaSignature` xml key is broken, we're getting in logs:
```
Winsparkle:
```
And that's it.

vslavik:
> But there is still one more issue worth considering: logging.

Done, thanks!

GitCop:

 * Commit: ea0a09554885d2b64ce41c4abaa6689bcdb92b13

 * Commit: 92f526a4378ff8f1f13d4675c7ded5965830883b

---

PR #160, opened 2018-01-17
Update CMake scripts

drizt:


cmake/CMakeLists.txt
```diff
@@ -76,19 +78,71 @@ set(PUBLIC_HEADERS
...
   ${ROOT_DIR}/include/winsparkle-version.h)
 
-set(RESOURCES winsparkle translations)
-
-# resource compilation for MinGW
-foreach(RESOURCE ${RESOURCES})
-  add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${RESOURCE}.o
-                     COMMAND ${CMAKE_RC_COMPILER} -I${CMAKE_CURRENT_SOURCE_DIR} -i${CMAKE_CURRENT_SOURCE_DIR}/${RESOURCE}.rc
-                             -o ${CMAKE_CURRENT_BINARY_DIR}/${RESOURCE}.o)
+set(LANGUAGES
```
vslavik: This is yet another place where languages list has to be maintained (and manually). Is there no way in CMake to look for all available *.po files instead?

cmake/common.rc
```diff
@@ -0,0 +1,7 @@
+/*
+MinGW can't compile multiple .rc files. Only single.
```
vslavik: Perhaps a better fix, simpler for MinGW, would be to include `translations.rc` from `winsparkle.rc` then?

cmake/expat/expat_config.h.cmake
```diff
@@ -0,0 +1,91 @@
+/* expat_config.h.in.  Generated from configure.in by autoheader.  */
```
vslavik: Why do you need this file at all, when it wasn't previously needed? This doesn't seem right, surely this won't be accurate for all possible compilers.

cmake/CMakeLists.txt
```diff
@@ -76,19 +78,71 @@ set(PUBLIC_HEADERS
...
+                   DEPENDS ${MO_FILES}
+)
+
+list(APPEND SOURCES ${CMAKE_CURRENT_BINARY_DIR}/common_res.o)
+
 add_library(${PROJECT_NAME} SHARED ${SOURCES} $<TARGET_OBJECTS:wxWidgets> $<TARGET_OBJECTS:expat>)
 
 target_link_libraries(${PROJECT_NAME} wininet version rpcrt4 comctl32)
 
+#set_target_properties(${PROJECT_NAME} PROPERTIES
```
vslavik: Please don't include commented out code — either it is needed, in which case it shouldn't be commented out, or it isn't, in which case it should be there at all.

vslavik (changes requested):

cmake/CMakeLists.txt
drizt: Fixed

cmake/CMakeLists.txt
puetzk: You could use `file(GLOB VARIABLE ${ROOT_DIR}/translations/*.po CONFIGURE_DEPENDS)` (https://cmake.org/cmake/help/v3.12/command/file.html#glob), but it was new in CMake 3.12 so you'd have to rause your cmake_minimum_required quite a bit. Before that there was no way to get CMake to re-configure if the glob results change. I'm not sure if 2.8 (which is your current minimum) even has file(GLOB) at all.

vslavik: @puetzk Thanks! How problematic is raising CMake version? I think not much on Windows where distro packaging doesn't enter the picture… AppVeyor, which I care about for CI, already includes 3.13.

vslavik: @puetzk However not using CMake myself, I don't really know how to translate your advice into actual code; a patch/PR would be tremendously helpful.

Youw: [file(GLOB)](https://cmake.org/cmake/help/v2.8.12/cmake.html#command:file) available at least since CMake 2.8, so no raising is required.

puetzk: But that's missing CONFIGURE_DEPENDS (which is what was new in 3.12), so it won't reconfigure properly if the filesystem changes.

Youw: So if and only if:
- already run cmake command (configure build/generate build scripts);
- after that you add some new _*.po_ file(s);
- and run cmake --build (or make, if this is mingw build, or build in Visual Studio);
In that case the build won't include new files.

If you run cmake (configure) command manually (which you normally do, if you have CI, or smth) - everything works.
If you remove existing .po file - build reasonably fails, until you run cmake.
If you do a clean build - everything works by default anyway.

Don't think this small feature is worth considering raising minimal required cmake version, specially for build system that is not officially supported.

I've seen two approaches, which I'd support widely:
1) Support cmake 2.8 and higher - great if you want to have(already have) a wide range of supported platforms/clients;
2) Support only latest cmake release - great for development, specially if you control all the build environments;

Anything in between sounds like not a great compromise.

PR #171, opened 2018-04-19
Failure case for Access Violation given fails_appcast.xml

mpleis:


vslavik:
The idea was to include a patch in a comment on the issue (see GitHub markup docs). Never mind, though, successfully reproduced.

PR #179, opened 2018-08-29
Allow manipulation of WinSparkle settings with user provided functions

BSG-75:
Added win_sparkle_set_config_methods(), which can be used to override WinSparkle's default Windows Registry interface with custom functions, so users can provide their own functions when they want to manage all configuration by themselves.

### DIFF
--- a/include/winsparkle.h
+++ b/include/winsparkle.h
@@ -248,7 +248,7 @@ typedef struct win_sparkle_config_methods_tag {
     int(__cdecl *config_read)(const char *name, wchar_t *buf, size_t len, void *user_data);
     /// Write @a value as config value @a name 's new value
     void(__cdecl *config_write)(const char *name, const wchar_t *value, void *user_data);
-    /// Delete config value named @ name
+    /// Delete config value named @a name
     void(__cdecl *config_delete)(const char *name, void *user_data);
     /// Arbitrary data which will be passed to the above functions, WinSparkle will not read or alter it.
     void *user_data;
--- a/include/winsparkle.h
+++ b/include/winsparkle.h
@@ -243,7 +243,7 @@ WIN_SPARKLE_API void __cdecl win_sparkle_set_app_build_version(const wchar_t *bu
 WIN_SPARKLE_API void __cdecl win_sparkle_set_registry_path(const char *path);
 
 /// Type used to override WinSparkle's config read function
-typedef int(__cdecl *win_sparkle_config_read_t)(void *custom_data, const char *name, wchar_t *buf, size_t bytes);
+typedef int(__cdecl *win_sparkle_config_read_t)(void *custom_data, const char *name, wchar_t *buf, size_t len);
 /// Type used to override WinSparkle's config write function
 typedef int(__cdecl *win_sparkle_config_write_t)(void *custom_data, const char *name, const wchar_t *value);
 /// Type used to override WinSparkle's config delete function
@@ -265,8 +265,8 @@ typedef int(__cdecl *win_sparkle_config_delete_t)(void *custom_data, const char 
     @param custom_data  arbitrary data which will be passed to your config manipulating 
                         functions, WinSparkle will not read or alter it.
 
-    @note The forth parameter *bytes* of function config_read is the size of *buf* in *bytes*, 
-          not array length!
+    @note There's no guarantee about the thread from which these functions are called,
+          Make sure your functions are thread-safe.
 
 */
 WIN_SPARKLE_API void __cdecl win_sparkle_set_config_methods(void *custom_data, 
--- a/include/winsparkle.h
+++ b/include/winsparkle.h
@@ -242,12 +242,18 @@ WIN_SPARKLE_API void __cdecl win_sparkle_set_app_build_version(const wchar_t *bu
  */
 WIN_SPARKLE_API void __cdecl win_sparkle_set_registry_path(const char *path);
 
-/// Type used to override WinSparkle's config read function
-typedef int(__cdecl *win_sparkle_config_read_t)(void *custom_data, const char *name, wchar_t *buf, size_t len);
-/// Type used to override WinSparkle's config write function
-typedef int(__cdecl *win_sparkle_config_write_t)(void *custom_data, const char *name, const wchar_t *value);
-/// Type used to override WinSparkle's config delete function
-typedef int(__cdecl *win_sparkle_config_delete_t)(void *custom_data, const char *name);
+/// Type used to override WinSparkle configuration's read, write and delete functions
+typedef struct win_sparkle_config_methods_tag {
+    /// Copy config value named @a name to the buffer pointed by @a buf, returns TRUE on success, FALSE on failure
+    int(__cdecl *config_read)(const char *name, wchar_t *buf, size_t len, void *user_data);
+    /// Write @a value as config value @a name 's new value
+    void(__cdecl *config_write)(const char *name, const wchar_t *value, void *user_data);
+    /// Delete config value named @ name
+    void(__cdecl *config_delete)(const char *name, void *user_data);
+    /// Arbitrary data which will be passed to the above functions, WinSparkle will not read or alter it.
+    void *user_data;
+} win_sparkle_config_methods_t;
+
 
 /**
     Override WinSparkle's configuration read, write and delete functions.
@@ -262,18 +268,15 @@ typedef int(__cdecl *win_sparkle_config_delete_t)(void *custom_data, const char 
     If you passed NULL as a configuration action (read, write or delete)'s function pointer, 
     WinSparkle will use the default function for that action.
 
-    @param custom_data  arbitrary data which will be passed to your config manipulating 
-                        functions, WinSparkle will not read or alter it.
+    @param config_methods  Your own configuration read, write and delete functions.
+                           Pass NULL to let WinSparkle continue to use its default functions.
 
     @note There's no guarantee about the thread from which these functions are called,
           Make sure your functions are thread-safe.
 
     @since 0.7
 */
-WIN_SPARKLE_API void __cdecl win_sparkle_set_config_methods(void *custom_data, 
-                                                            win_sparkle_config_read_t   config_read,
-                                                            win_sparkle_config_write_t  config_write,
-                                                            win_sparkle_config_delete_t config_delete);
+WIN_SPARKLE_API void __cdecl win_sparkle_set_config_methods(win_sparkle_config_methods_t *config_methods);
 
 /**
     Sets whether updates are checked automatically or only through a manual call.
--- a/include/winsparkle.h
+++ b/include/winsparkle.h
@@ -254,7 +254,7 @@ typedef int(__cdecl *win_sparkle_config_delete_t)(void *custom_data, const char 
 
     By default, WinSparkle will read, write and delete configuration values by
     interacting directly with Windows Registry.
-    If you want to manage configuration by yourself or you don't want let WinSparkle 
+    If you want to manage configuration by yourself, or if you don't want let WinSparkle 
     write settings directly to the Windows Registry, you can provide your own functions 
     to read, write and delete configuration.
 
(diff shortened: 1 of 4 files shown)


GitCop:

 * Commit: 9dab2e4c68cc3f1d7b4965d6bc0701331446a364

---

include/winsparkle.h
```diff
@@ -242,6 +242,38 @@ WIN_SPARKLE_API void __cdecl win_sparkle_set_app_build_version(const wchar_t *bu
...
+    to read, write and delete configuration.
+
+    These functions needs to return TRUE on success, FALSE on failure.
+    If you passed NULL as a configuration action (read, write or delete)'s function pointer, 
+    WinSparkle will use the default function for that action.
+
+    @param custom_data  arbitrary data which will be passed to your config manipulating 
+                        functions, WinSparkle will not read or alter it.
+
+    @note The forth parameter *bytes* of function config_read is the size of *buf* in *bytes*,
```
Youw: this would confuce so many users
there is no reasons, why this shouldn't be an array length

include/winsparkle.h
```diff
@@ -242,6 +242,38 @@ WIN_SPARKLE_API void __cdecl win_sparkle_set_app_build_version(const wchar_t *bu
  */
 WIN_SPARKLE_API void __cdecl win_sparkle_set_registry_path(const char *path);
 
+/// Type used to override WinSparkle's config read function
+typedef int(__cdecl *win_sparkle_config_read_t)(void *custom_data, const char *name, wchar_t *buf, size_t bytes);
+/// Type used to override WinSparkle's config write function
+typedef int(__cdecl *win_sparkle_config_write_t)(void *custom_data, const char *name, const wchar_t *value);
```
Youw: inconsistent naming
`buf` and `value` are the same entity
is there a reason why you would name them differently?

include/winsparkle.h
BSG-75: ...Because the original WinSparkle `RegistryRead(name, buf, sizeof(buf))` use the number of bytes (as you can see from `sizeof(buf)`), and I've left it unchanged.
But I think you're right, I'll change the code so it uses array length.

include/winsparkle.h
vslavik: @Youw That's clearly copied from the existing code, and I wouldn't consider it a problem: the read function *is* reading into a buffer and if anything, it emphasizes the need to specify the buffer's length.

BSG-75: Hmmm... Because it doesn't make sense that users needs to "provide a value" when they need to read the value - which means they don't have a value yet, they can only provide a buffer to receive value.

include/winsparkle.h
vslavik: I don't really understand what you mean by "array length" and how it differs from "buffer size"? The code this is derived from uses buffer length as passed to win32 functions and is the standard way to declare how much space is available for writing...

BSG-75: If we have a `wchar_t buf[512]`, the array length is 512, but the buffer size is 1024. And in existing codes WinSparkle pass 1024 instead of 512, so I decided to keep it unchanged and follow this convention.
But I agree it's pretty counterintuitive, I've added that "it's buffer size in bytes, not array length!" Exactly because I have experienced crashes due memory corruption...

Youw: Imagine next example:
```cpp
int __cdecl config_read_impl(void *custom_data, const char *name, wchar_t *buf, size_t bytes)
{
    std::wstring the_value("This is some very long (or not) value - got from our reader"); // 60+1 character
    size_t max_len_to_write = std::max(the_value.size(), bytes-1);
    wcsncpy(buf, the_value.c_str(), max_len_to_write);
    buf[max_len_to_write] = L'\0';
    return TRUE;
}
```
and the fact, that `sizeof(wchar_t) == 2`
so many users will find a memory corruption here or broken values

despite the fact, that usage is not as per documentation, but this way is much more expected

include/winsparkle.h
Youw: maybe you're right
just my own observation - fresh look without a context

BSG-75:
I've just noted that the majority of WinSparkle API has their minimum version requirement in the API description, from example `@since 0.3` or `@since 0.4`.
Do I have to add a version requirement for `win_sparkle_set_config_methods` as well, or you will take care of that?

include/winsparkle.h
BSG-75: I've updated the code, now the function accepts array length instead of size in bytes.

src/settings.h
```diff
@@ -217,6 +218,35 @@ class Settings
...
+    static void SetConfigMethods(void *customConfigData, 
+                                win_sparkle_config_read_t customConfigRead,
+                                win_sparkle_config_write_t customConfigWrite, 
+                                win_sparkle_config_delete_t customConfigDelete)
+    {
+        CriticalSectionLocker lock(ms_csVars);
+        
+        ms_currentConfigRead = DefaultConfigRead;
+        ms_currentConfigWrite = DefaultConfigWrite;
+        ms_currentConfigDelete = DefaultConfigDelete;
```
Youw: So now there is a small amount of time between this line and the ones below, when `ms_currentConfig*` may be `nullptr`.
To avoid the most unreal scenario, when one of `Do*ConfigValue` functions called in exactly this moment, I suggest either lock `g_csConfigValues` in this function as well, either move these assignments under `else` below.

Youw:
Looks good to me.

vslavik:
> Do I have to add a version requirement for win_sparkle_set_config_methods by myself as well (but I don't know what's the next version number)

Ideally a PR would be self-contained and direct-mergeable, so it would be best to have them — for an API addition, the version will need to increase to 0.7.

include/winsparkle.h
```diff
@@ -242,6 +242,38 @@ WIN_SPARKLE_API void __cdecl win_sparkle_set_app_build_version(const wchar_t *bu
...
+    to read, write and delete configuration.
+
+    These functions needs to return TRUE on success, FALSE on failure.
+    If you passed NULL as a configuration action (read, write or delete)'s function pointer, 
+    WinSparkle will use the default function for that action.
+
+    @param custom_data  arbitrary data which will be passed to your config manipulating 
+                        functions, WinSparkle will not read or alter it.
+
+    @note There's no guarantee about the thread from which these functions are called,
```
vslavik: The calling code *does* provide some guarantees though: namely that all calls to the config functions are serialized (by means of `g_csConfigValues`) — i.e. you don't have to worry about thread-safely in the implementation most of the time, i.e. unless you use thread local storage for something.

src/settings.cpp
```diff
@@ -339,7 +368,12 @@ void Settings::DeleteConfigValue(const char *name)
 {
     CriticalSectionLocker lock(g_csConfigValues);
 
-    RegistryDelete(name);
+    if (0 == ms_currentConfigDelete(ms_customConfigData, name))
```
vslavik: Please use either boolean comparison (`if (!…)`) or with the constant at the end (`… == 0)`), as a human would say it (other places too).

BSG-75:
I've just added the `@since 0.7` to `win_sparkle_set_config_methods`'s API description.

src/settings.cpp
```diff
@@ -44,6 +44,10 @@ std::wstring Settings::ms_appVersion;
 std::wstring Settings::ms_appBuildVersion;
 std::string Settings::ms_DSAPubKey;
 
+void *Settings::ms_customConfigData = nullptr;
+win_sparkle_config_read_t Settings::ms_currentConfigRead = Settings::DefaultConfigRead;
```
vslavik: Missing `&` in front of the assigned value (`… = &Settings:…`) unless I'm mistaken.

src/settings.h
```diff
@@ -323,6 +359,11 @@ class Settings
     static std::wstring ms_appVersion;
     static std::wstring ms_appBuildVersion;
     static std::string  ms_DSAPubKey;
+    
+    static void *ms_customConfigData;
+    static win_sparkle_config_read_t ms_currentConfigRead;
```
vslavik: I don't think this needs to say "current" in it — it's not like it's something that is supposed to change during lifetime;   OTOH it would make sense to indicate it's a function, because it's not entirely obvious when reading the code elsewhere, so `ms_configReadFunc, …, m_configCustomData`.

src/settings.cpp
```diff
@@ -316,20 +320,45 @@ CriticalSection g_csConfigValues;
...
 
-void Settings::DoWriteConfigValue(const char *name, const wchar_t *value)
+int __cdecl Settings::DefaultConfigRead(void *, const char *name, wchar_t* buf, size_t len)
 {
-    CriticalSectionLocker lock(g_csConfigValues);
+    //RegistryRead requires array size in bytes
+    return RegistryRead(name, buf, len * sizeof(wchar_t));
+}
 
+int __cdecl Settings::DefaultConfigWrite(void *, const char *name, const wchar_t* value)
```
vslavik: Actually the more important comment here is that these very thin wrappers don't serve a useful role — they just trivially call another *private* function, so let's just refactor the private functions to be directly usable.

src/settings.cpp
```diff
@@ -316,20 +320,45 @@ CriticalSection g_csConfigValues;
...
+
+void Settings::DoWriteConfigValue(const char *name, const wchar_t *value)
+{
+    CriticalSectionLocker lock(g_csConfigValues);
+    
+    if (0 == ms_currentConfigWrite(ms_customConfigData, name, value))
+    {
+        //Only custom write function will return 0, 
+        //default function would already have thrown an exception.
+        throw std::runtime_error("User provided config write function failed");
```
vslavik: I don't see value in throwing like this — let the user-provided function do its own error handling. It's more graceful when WinSparkle silently ignores write failures, as it does now, then when it throws errors (and possibly crashes or spams with errors, the host application).

There's similarly little value in the return values if they're not actionable (although it might make sense to have them as a sort of future-proofing). It's useful in the *Read* function, which needs to fail gracefully in absence of config data, but the other are mostly ignorable.

src/settings.h
```diff
@@ -217,6 +218,35 @@ class Settings
...
+    {
+        CriticalSectionLocker lock(ms_csVars);
+        
+        ms_currentConfigRead = DefaultConfigRead;
+        ms_currentConfigWrite = DefaultConfigWrite;
+        ms_currentConfigDelete = DefaultConfigDelete;
+
+        ms_customConfigData = customConfigData;
+
+        if (customConfigWrite)
```
vslavik: This is dangerous. It allows overriding only some of the functions, but not others — I think this should be all-or-nothing: if you customize config functions, you need to customize all of them.

(An alternative API to drive the point home would be to use not 3 separate arguments, but a "virtual table-like" struct with function pointers.)

include/winsparkle.h
BSG-75: I was mainly concerning about the data race between WinSparkle's config functions and application's config functions.

For example if application stores all its configuration (including WinSparkle's settings) inside a single json file, then if application's main thread and WinSparkle's thread write contents to the file at same time, there will be a data race.

BSG-75:
By the way I have a question that's not directly related to this pull request: does WinSparkle use C++11 (or even C++17)? Because I've seen `nullptr` in WinSparkle's code, but on the readme file it's said that "projects for Visual C++ (2008 and up) are provided", and I doubt VS2008 can compile WinSparkle if it uses C++11 code..

vslavik:
> By the way I have a question that's not directly related to this pull request: does WinSparkle use C++11 (or even C++17)?

It's probably still better not to (that `nullptr` you've seen crept in by mistake), but OTOH, apparently nobody builds with old compilers anymore. The moment C++11 is useful for anything in the code, I wouldn't hesitate to start requiring it.

BSG-75:
> > By the way I have a question that's not directly related to this pull request: does WinSparkle use C++11 (or even C++17)?
> 
> It's probably still better not to (that `nullptr` you've seen crept in by mistake), but OTOH, apparently nobody builds with old compilers anymore. The moment C++11 is useful for anything in the code, I wouldn't hesitate to start requiring it.

I'm planning to add delta updates support for WinSparkle, and then I found those long chain of `if`/`else` in the Xml parsing code of `appcast.cpp`. Without wanting to make that chain of `if` and `else`s longer, I think by using `std::function` and lambdas (those two needs C++11) together with `std::map` can greatly simplify those code.

vslavik:
> I think by using std::function and lambdas (those two needs C++11) together with std::map can greatly simplify those code.

That… seems doubtful. This part is terrible, no argument there, but the only sensible fix — which I'm planning on doing — is to tear it out and replace with a nice DOM parsing API (pugixml or tinyxml2… I don't think anybody would characterize using Windows' COM-based parser API as "nice").

vslavik:
Merged in c5f8d27.

PR #181, opened 2018-09-08
Add support of custom http headers.

ostrobrod:
Here is implemented setup of custom HTTP headers.

Two functions were added in API:
`win_sparkle_add_http_header` adds new header parameter.
`win_sparkle_clear_http_headers` clear all custom header parameters.

Header storage is implemented as `std::wstring`.

### DIFF
--- a/include/winsparkle-version.h
+++ b/include/winsparkle-version.h
@@ -31,7 +31,7 @@
  *--------------------------------------------------------------------------*/
 
 #define WIN_SPARKLE_VERSION_MAJOR   0
-#define WIN_SPARKLE_VERSION_MINOR   6
+#define WIN_SPARKLE_VERSION_MINOR   7
 #define WIN_SPARKLE_VERSION_MICRO   0
 
 /**
--- a/include/winsparkle-version.h
+++ b/include/winsparkle-version.h
@@ -31,7 +31,7 @@
  *--------------------------------------------------------------------------*/
 
 #define WIN_SPARKLE_VERSION_MAJOR   0
-#define WIN_SPARKLE_VERSION_MINOR   7
+#define WIN_SPARKLE_VERSION_MINOR   6
 #define WIN_SPARKLE_VERSION_MICRO   0
 
 /**
--- a/src/download.cpp
+++ b/src/download.cpp
@@ -188,7 +188,7 @@ void WaitUntilSignaledWithTerminationCheck(Event& event, Thread *thread)
                                 public functions
  *--------------------------------------------------------------------------*/
 
-void DownloadFile(const std::string& url, IDownloadSink *sink, Thread *onThread, int flags)
+void DownloadFile(const std::string& url, IDownloadSink *sink, Thread *onThread, const std::string &headers, int flags)
 {
     char url_path[2048];
     URL_COMPONENTSA urlc;
@@ -234,8 +234,8 @@ void DownloadFile(const std::string& url, IDownloadSink *sink, Thread *onThread,
                          (
                              inet,
                              url.c_str(),
-                             NULL, // lpszHeaders
-                             -1,   // dwHeadersLength
+							 headers.c_str(), // lpszHeaders
+                             (DWORD)headers.length(),   // dwHeadersLength
                              dwFlags,
                              (DWORD_PTR)&context  // dwContext
                          );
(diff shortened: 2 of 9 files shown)


include/winsparkle-version.h
```diff
@@ -31,7 +31,7 @@
  *--------------------------------------------------------------------------*/
 
 #define WIN_SPARKLE_VERSION_MAJOR   0
-#define WIN_SPARKLE_VERSION_MINOR   6
+#define WIN_SPARKLE_VERSION_MINOR   7
```
vslavik: This is unrelated to the rest and shouldn't be included in this commit (or, in fact, this PR at all). Please follow the "one commit per one logical thing" rule of git use.

src/download.cpp
```diff
@@ -234,8 +234,8 @@ void DownloadFile(const std::string& url, IDownloadSink *sink, Thread *onThread,
                          (
                              inet,
                              url.c_str(),
-                             NULL, // lpszHeaders
-                             -1,   // dwHeadersLength
+							 headers.c_str(), // lpszHeaders
```
vslavik: As you can see yourself here, your changes break formatting all over the place. Please respect existing code's indentation etc. style, including the use of tabs vs. spaces and the amount of indentation.

src/updatechecker.cpp
```diff
@@ -228,7 +228,7 @@ void UpdateChecker::PerformUpdateCheck()
         CheckForInsecureURL(url, "appcast feed");
 
         StringDownloadSink appcast_xml;
-        DownloadFile(url, &appcast_xml, this, Download_BypassProxies);
+        DownloadFile(url, &appcast_xml, this, WideToAnsi(Settings::GetHttpHeaderString()), Download_BypassProxies);
```
vslavik: What's the point of using wide strings for the API and storage, when you're going to throw it all away (without even a valid conversion to something like UTF-8)? It doesn't make sense for HTTP headers anyway.

vslavik:
Thanks! I have questions, though:

1. Why?
2. Does Sparkle (for Mac) has anything similar?

At first glance, this seems extremely specialized to be worth making part of the API and maybe a more generic approach with customizing the HTTP request object would make more sense -- but maybe I'm missing something.

src/settings.h
```diff
@@ -323,6 +345,7 @@ class Settings
     static std::wstring ms_appVersion;
     static std::wstring ms_appBuildVersion;
     static std::string  ms_DSAPubKey;
+	static std::string  ms_httpHeaders;
```
Youw: tab again

src/updatedownloader.cpp
```diff
@@ -177,7 +177,7 @@ void UpdateDownloader::Run()
       Settings::WriteConfigValue("UpdateTempDir", tmpdir);
 
       UpdateDownloadSink sink(*this, tmpdir);
-      DownloadFile(m_appcast.DownloadURL, &sink, this);
+      DownloadFile(m_appcast.DownloadURL, &sink, this, "");
```
Youw: maybe a default value for this parameter instead?

src/settings.h
```diff
@@ -196,6 +197,27 @@ class Settings
         ms_appVersion = version;
     }
 
+	/// Adds an additional header parameter.
+	static void AddHttpHeader(const char *name, const char *value)
+	{
+		CriticalSectionLocker lock(ms_csVars);
+		ms_httpHeaders += std::string(name) + std::string(": ") + std::string(value) + std::string("\r\n");
```
Youw: I don't believe, creating temporary objects is necessary here:
for `std::string(": ")` and `std::string("\r\n")` SSO will do its job, but for `std::string(value)` it will most likely lead to memory allocation and then deallocation

vslavik: More importantly, and avoiding this issue, it would make more sense to have `ms_httpHeaders` as `map<string, string>`, to allow overriding and to make duplicate headers impossible...

vslavik:
> In my case, I need send some additional information about a user to the server (this information is used for licensing).

Ah, that makes a lot of sense, thanks for the explanation!

vslavik:
Thanks, merged in 506c0b21b35b9f524e775344327bbd1e3a484d02 and 0c34819f2770e273acd34b1c6634e4d2bdaf0ca5 after some slight cleanup (and small API change, to use `win_sparkle_set_http_header()` instead of "add", in another commit).

PR #186, opened 2018-12-05
added allow downgrade functionality

arobertsdisp:
Some organizations may need the ability to allow application downgrades. This pull request adds a new config value that can be set to allow downgrades. If downgrades are allowed, we only check to make sure the appcast version is different than the current version. If downgrades are not allowed, we check to make sure the appcast version is newer than the current version.

### DIFF
--- a/include/winsparkle.h
+++ b/include/winsparkle.h
@@ -397,6 +397,26 @@ typedef void (__cdecl *win_sparkle_update_cancelled_callback_t)();
 */
 WIN_SPARKLE_API void __cdecl win_sparkle_set_update_cancelled_callback(win_sparkle_update_cancelled_callback_t callback);
 
+/**
+Gets the allow downgrade state
+
+@return  1 if downgrades are allowed, 0 otherwise
+
+@note Defaults to 0 when not yet configured (as happens on first start).
+
+@since 0.6.1
+*/
+WIN_SPARKLE_API int __cdecl win_sparkle_get_allow_downgrades();
+
+/**
+Sets the allow downgrade state.
+
+@param  allowDowngrade 1 to allow downgrades, 0 otherwise
+
+@since 0.6.1
+*/
+WIN_SPARKLE_API void __cdecl win_sparkle_set_allow_downgrades(int allowDowngrade);
+
 //@}
 
 
--- a/src/dll_api.cpp
+++ b/src/dll_api.cpp
@@ -315,6 +315,28 @@ WIN_SPARKLE_API void __cdecl win_sparkle_set_update_cancelled_callback(win_spark
     CATCH_ALL_EXCEPTIONS
 }
 
+WIN_SPARKLE_API void __cdecl win_sparkle_set_allow_downgrades(int allowDowngrades)
+{
+	try
+	{
+		Settings::WriteConfigValue("AllowDowngrades", allowDowngrades != 0);
+	}
+	CATCH_ALL_EXCEPTIONS
+}
+
+WIN_SPARKLE_API int __cdecl win_sparkle_get_allow_downgrades()
+{
+	try
+	{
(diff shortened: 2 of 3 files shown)


vslavik:
As done here, this is terribly insecure — it allows anybody to enable this insecure behavior in any application using WinSparkle by just adding a registry key. Sparkle intentionally doesn't allow downgrades, and I agree with their rationale. I won't be merging this PR, sorry.

I'm willing to consider a hidden *compile-time* flag enabling this, [similarly to how Sparkle does it](https://github.com/sparkle-project/Sparkle/search?q=SPARKLE_AUTOMATED_DOWNGRADES&type=Code). If you'd like to pursue that route, please submit a new PR; please make the commit's description [follow best practices guidelines](
http://chris.beams.io/posts/git-commit/) too.

arobertsdisp:
Would you consider a PR that made use of a callback to determine whether downgrades are allowed? This would be more secure as the allowing of downgrades would need to be compiled into your application and  WinSparkle would simply check the value using a callback.

vslavik:
Sorry, but no. This is something that IMNSHO shouldn't be there at all; the most I'm willing to consider is a *minimalistic,* minimally intrusive patch that I can mostly ignore (see the Sparkle link I provided) and that doesn't add any significant code to be maintained.

arobertsdisp:
IMNSHO downgrades are not insecure, especially when being deployed through Sparkle/WinSparkle. For someone with bad intentions to break into the system they would need to

1. Get access to the https server where the appcast.xml lives and modify the version to be an older version
2. Somehow modify the application code to change the callback to return a value that indicates downgrades are allowed.

Assuming somehow someone was able to this, step #1 means the entire system has been compromised as they could modify the appcast to download any installer they wish (possibly containing malicious code). This has nothing to do with downgrades being more insecure than upgrades.

So, you argument about downgrades being insecure is weak but this is your repository so I understand the final decision is yours. I suspect there are others out there that would like this functionality. Stuff happens/accidents are made and sometimes updates go out that shouldn't have. It would be nice to be able to remedy this situation in those cases.

vslavik:
> Somehow modify the application code to change the callback to return a value 

That objection was against the terrible idea of storing it as a registry value. The objection against using a callback is that it makes no damn sense to *set a value* via a callback and adds overcomplicated code that somebody has to maintain. 

> Stuff happens/accidents are made and sometimes updates go out that shouldn't have. It would be nice to be able to remedy this situation in those cases.

Releasing a *new* version is the traditional way to do that...


The only form in which I'm willing to allow this functionality is as a compile-time toggle, as Sparkle does. I'm making this call as the maintainer. Let's please not waste any more time arguing about it.

arobertsdisp:
>  it allows anybody to enable this insecure behavior

Your original response indicated you thought downgrades were "insecure behavior" and that the method of turning this on (registry key) was also insecure.

>  The objection against using a callback is that it makes no damn sense to set a value via a callback and adds overcomplicated code that somebody has to maintain.

You are currently using a callback to control whether WinSparkle can shutdown the app before starting the upgrade. I don't see how using a callback in the case of downgrades would be any different.

FWIW, if you don't want to waste any more time, you can always not respond.

vslavik:
Thanks for being so respectful of maintainer's wishes.

> I don't see how using a callback in the case of downgrades would be any different.

Yes, it's apparent that you don't see the implications of your changes. W.r.t. callbacks, one is for determining a property of the hosting process that is dynamic, changes in time, and must be known at a time determined by WinSparkle. The other is a bizarrely convoluted way of setting a boolean flag that is constant and determined solely by the hosting process. IOW, "let's make it a callback!" is not a reasonable solution to "you can't allow this boolean to be settable externally" — *removing*, not adding, code would be.

Issue #138, opened 2017-06-07
Not able to build Winsparkle code


kguduri:
translations\translations.rc(1): error RC2135: file not found: ar.mo

all the translations file are not available adding the screen shot for the same, Followed the same steps that are mentioned in "Building from Sources"

![image](https://user-images.githubusercontent.com/27999771/26873811-c807aedc-4b98-11e7-8eec-022335d12ec4.png)

Please help me on this as soon as possible.

Or is there any other way where i can remove/hide the "Skip this version".

vslavik:
> as soon as possible.

I didn't realize you were paying for support, so maybe tone it down a notch...

If you want fast help, you should start with actually describing what you did in meaningful detail, such as at least mentioning basics like the exact compiler version used or providing full (verbose!) compilation log.

The little that is visible in the censored screenshot indicates that the normal build process was *not* run, because the targets from `WinSparkle.Targets` were *not* executed. I don't see how that's possible unless you modified something.

Enable build logging, investigate.

kguduri:
I have done the following to get the latest code 
$ git clone git://github.com/vslavik/winsparkle.git
$ cd winsparkle
$ git submodule init
$ git submodule update

Opened WinSparkle.sln in visual stuido 2015 and tried building the solution but the build fail with the following messages
translations\translations.rc(1): error RC2135: file not found: ar.mo
3>
3>
3>translations\translations.rc(2): error RC2135: file not found: be.mo
3>
3>
3>translations\translations.rc(3): error RC2135: file not found: bg.mo
3>
3>
3>translations\translations.rc(4): error RC2135: file not found: bs.mo
3>
3>
3>translations\translations.rc(5): error RC2135: file not found: ca_ES.mo
3>
3>
3>translations\translations.rc(6): error RC2135: file not found: co.mo
3>
3>
3>translations\translations.rc(7): error RC2135: file not found: cs.mo
3>
3>
3>translations\translations.rc(8): error RC2135: file not found: da.mo
3>
3>
3>translations\translations.rc(9): error RC2135: file not found: de.mo
3>
3>
3>translations\translations.rc(10): error RC2135: file not found: el.mo
3>
3>
3>translations\translations.rc(11): error RC2135: file not found: es.mo
3>
3>
3>translations\translations.rc(12): error RC2135: file not found: eu.mo
3>
3>
3>translations\translations.rc(13): error RC2135: file not found: fr.mo
3>
3>
[...]

PR #197, opened 2019-04-26
remove redundant explicit pointer initializations which are incompatible with older c++ versions

matthiesenj:
This is necessary to build in e.g. VS2012. #138

### DIFF
--- a/src/signatureverifier.cpp
+++ b/src/signatureverifier.cpp
@@ -210,8 +210,8 @@ private:
 
     public:
         BIOWrap(const std::string &mem_buf)
+            : bio(BIO_new_mem_buf(mem_buf.c_str(), int(mem_buf.size())))
         {
-            bio = BIO_new_mem_buf(mem_buf.c_str(), int(mem_buf.size()));
             if (!bio)
                 throw std::invalid_argument("Cannot set PEM key mem buffer");
         }
@@ -238,6 +238,7 @@ public:
 
     public:
         DSAPub(const std::string &pem_key)
+            : dsa(NULL)
         {
             BIOWrap bio(pem_key);
             if (!PEM_read_bio_DSA_PUBKEY(bio, &dsa, NULL, NULL))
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: 993f1f8382038cfa9629d9e8912b9bc0361fd0c3

---

Youw:
I'd add an initialization in c-tor instead.

matthiesenj:
> I'd add an initialization in c-tor instead.

They are already initialized in their respective ctors, so the explicit member initializations are redundant.

Youw:
Those are not _initialized_.
In BIOWrap it is _assigned_, before first usage, which, strictly, fine, but considered a bad practice in C++ world.
In DSAPub the variable is passed to some 3d-party function _uninitialised_, which may lead to undefined behavior under some circumstances.

matthiesenj:
Ok, wrt `PEM_read_bio_DSA_PUBKEY` it seems you are right there could be a problem since the function does different things depending on if `dsa is `NULL` or not.

matthiesenj:
@Youw better now?

vslavik:
Thanks!

How much do you depend on using an old compiler like 2012? I hoped to start using more of modern C++, not less.

vslavik:
Squashed in 798d11841915952f35afb17d0500460cf8feba3e together with the linkage fix you described in the other issue. Thanks!

matthiesenj:
> 
> 
> Thanks!
> 
> How much do you depend on using an old compiler like 2012? I hoped to start using more of modern C++, not less.

I am just about to switch to VS2019, so in that respect, supporting old  compilers is not a problem.
The project clearly indicates itself (through the project files) as being buildable in older VS versions, and that I think is a problem, if that is actually not the case. That's why I mentioned in #138 that versions which are no longer maintained maybe should just be branched off, so it is clear, which VS versions are supported by the current configuration of the code :-)

Issue #196, opened 2019-04-25
Error callback not invoked if error during automatic update check at application start


matthiesenj:
It would be nice to get notified that there is a problem with automatic checking, also when no ui (related to the current check!) is yet showing (to e.g. log a warning somewhere).
Would a sensible solution be to call `ApplicationController::NotifyUpdateError()` in `UI:NotifyUpdateError` when the ui thread is not running?

something like this:

```
void UI::NotifyUpdateError(ErrorCode err)
{
    UIThreadAccess uit;

    if ( !uit.IsRunning() )
    {
        ApplicationController::NotifyUpdateError();
        return;
    }

    ...
}
```

vslavik:
Let me be more radical: I think `UI::NotifyUpdateError()` should *always*, unconditionally, call `ApplicationController::NotifyUpdateError()` as the first thing it does, consistently with `UI:: NotifyNoUpdates` or `NotifyUpdateAvailable ` methods (and `m_errorOccurred` shouldn't exist anymore and nothing else should call it). I'd appreciate a PR to that effect.

Unless I'm missing something? I don't understand why the notifications work the way they do now, which is to wait for user's acknowledgment of the error (if UI is visible) before notifying. Does anybody?

matthiesenj:
Alright. That would be a much cleaner solution, yes. My suggestion was based on, that there was a sensible reason for the current design (with `m_errorOccurred` and whatnot) and trying not to break it :-)

As far as I can immediately see,  it looks safe to remove `m_errorOccurred` and related business, and just always invoke the callback in `UI:NotifyUpdateError`.

PR #198, opened 2019-04-26
Notify about update errors with/without ui showing

matthiesenj:
#196 Did not remove m_errorOccurred as it is used to guard against the ui calling `ApplicationController::NotifyUpdateCancelled()`.


vslavik:
Thanks!

PR #212, opened 2020-06-10
Add Visual Studio 2019 Files

Amadeus-:


vslavik:
Thanks. How did you create them? (Trying to determine what the differences will be.)

Amadeus-:
I just copied the 2017 files and then manually edited them to use the Visual Studio 2019 PlatformToolset (v142).   That's really the only changes necessary from 2017 to 2019.    To be honest though, you could just have one set of project files for 2015-2019+ ...since Visual Studio 2017 and 2019 will open projects and compile them using any of the PlatformToolsets from 2015 onward.   So -- the only difference between the 2017 and 2019 project files is that the 2019 files require having Visual Studio 2019.

I tested them as well.

Also, as a sidebar, Visual Studio 2019 PlatformToolset (v142) drops support for Windows XP.   Visual studio 2015, for example, had v140 and v140_xp ...as did Visual Studio 2017 (v141 and v141_xp) ....but, Visual Studio 2019 does not have an _xp option.

vslavik:
Merged, thanks!

vslavik:
> To be honest though, you could just have one set of project files for 2015-2019+ 

That's indeed the plan, but there are issues: 

> ...since Visual Studio 2017 and 2019 will open projects and compile them using any of the PlatformToolsets from 2015 onward.

But it needs to have the optional older-version toolsets installed...

> Visual Studio 2019 PlatformToolset (v142) drops support for Windows XP

And this is an issue too, although that one probably just means that official `WinSparkle.dll` still needs to be compiled with 2017.

PR #218, opened 2020-08-16
Add more notification callbacks to the API

systemdoni:


vslavik:
Can you pleas provide at least some explanation and rationale? I don't see much reason for such unimportant callbacks, so am reluctant to do it. Understanding the *why* may, and often does, change that. 

Please use [explanatory commit messages](https://chris.beams.io/posts/git-commit/) — from something like "Change calling NotifyUpdateDismissed() from OnCloseButton()" it's impossible to understand *why* it does what it does, and I can't apply a change with no reason behind it, nor how is it related to the subject of this PR. Similarly, a PR with a non-described "add new API" title and no description provided is hard to asses — same rationale as for having good commit messages applies to PR descriptions.

TIA!

systemdoni:
I am developing an App where I needed these callbacks. When I developed the macOS version, Sparkle had these callbacks meanwhile, WinSparkle didn't, so I added them for my project. I thought maybe someone else may need them.

Thanks for the info about explanatory commit messages. I will keep that in mind for future commits.
The last commit was actually a fix to make the function's behavior similar to Sparkle framework, so a notification will be sent only when the update dialog is closed from the close button. So instead of calling NotifyUpdateDismissed from onClose, I moved it to onCloseButton.

I hope this explanation is clearer.

include/winsparkle.h
```diff
@@ -483,6 +483,22 @@ typedef void(__cdecl* win_sparkle_update_postponed_callback_t)();
...
 
+/// Callback type for win_sparkle_update_dismissed_callback()
+typedef void(__cdecl* win_sparkle_update_dismissed_callback_t)();
+
+/**
+    Set callback to be called when the user dismisses
+    (closes) update dialog.
+
+    This is useful in combination with
+    win_sparkle_check_update_with_ui() or similar as it
```
vslavik, line 494: What is the "or similar" referring to?

vslavik:
> I am developing an App where I needed these callbacks.

I mean, if you didn't feel the need, you wouldn't need them — that's evident. But could you explain *why* you need them? Again, it's always good if maintainers understand what an API *is used for*, otherwise they risk screwing things up unwittingly. 

> Sparkle had these callbacks 

OK, that's a valid point.

> so a notification will be sent only when the update dialog is closed from the close button. So instead of calling NotifyUpdateDismissed from onClose, I moved it to onCloseButton.

I can see that — that's just saying in words what the code says already, but it doesn't say *why* and *what effect* does this change have. But since it's just a follow-up to the previous commit, I guess the actual question is why is it two commits instead of an atomic one? I take it it's just a work-in-progress update that can be squashed together, then?

include/winsparkle.h
systemdoni: functions similar to it, like win_sparkle_check_update_with_ui_and_install()

systemdoni:
> I mean, if you didn't feel the need, you wouldn't need them — that's evident. But could you explain _why_ you need them? 
Again, it's always good if maintainers understand what an API _is used for_, otherwise they risk screwing things up unwittingly.

It is a launcher for another app, so before launching the other app, I had to know what does the user chooses.

> I can see that — that's just saying in words what the code says already, but it doesn't say _why_ and _what effect_ does this change have. But since it's just a follow-up to the previous commit, I guess the actual question is why is it two commits instead of an atomic one? I take it it's just a work-in-progress update that can be squashed together, then?

Yes, I should have squashed the two commits together, my bad.

systemdoni:
I squashed the two last commits together
Regarding the callbacks I added, I couldn't think of any other case where they would be used, except my project. They may not be used at all, so maybe you should close this PR.

vslavik:
> Regarding the callbacks I added, I couldn't think of any other case where they would be used, except my project. They may not be used at all, so maybe you should close this PR.

Nah, I'm actually convinced by your explanation — it makes sense, is useful to you, is consistent with Sparkle.

Merged now, thanks a lot!

PR #220, opened 2020-10-22
Added callback to handle zip files in app instead of expecting an installer to run

patrickkidd:
WinSparkle expects the download to be an executable installer. This is a problem when the download is a zip file and merely contains an updated copy of the executable (i.e. for statically linked executables with no dependencies or installer). This adds a callback so that zip files can be handled in the app.

This pull request is not very clean. It doesn't add anything to the api but does expose the new function in the api. Maybe you can think of a better way to add this cleanly. But there should be some way to handle download payloads that are not installers.


translations/translations.rc
```diff
@@ -1,41 +0,0 @@
-winsparkle_af      MOFILE  "af.mo"
```
vslavik, line 1: This does not "Fix compile errors with translations", this completely breaks the translations.

.gitignore
```diff
@@ -20,3 +20,10 @@ packages
 pascal/lib/
 pascal/backup/
 .vs
+
+# pks
+3rdparty
+Backup/
+WinSparkle.vcxproj
+*.filters
+examples/example_psdk.vcxproj
```
vslavik, line 29: This does not ignore "build artifacts", but adds, wrongly, many files *that are committed to the repository* (which never makes any sense), ignores requires submodules (ditto) and adds some private files of yours (use your local git config for that).

src/dll_api.cpp
```diff
@@ -414,4 +414,9 @@ WIN_SPARKLE_API void __cdecl win_sparkle_check_update_without_ui()
 }
 
 
+WIN_SPARKLE_API void __cdecl win_sparkle_set_zip_download_callback(void (*cb)(std::wstring))
```
vslavik, line 417: Contrary to the claim that this adds nothing to the API, you're doing just that here — but without adding it to the public header or documenting it.

src/ui.cpp
```diff
@@ -696,8 +696,21 @@ void UpdateDialog::OnRunInstaller(wxCommandEvent&)
     }
 }
 
+static inline bool ends_with(std::wstring const& value, std::wstring const& ending)
+{
+    if (ending.size() > value.size()) return false;
+    return std::equal(ending.rbegin(), ending.rend(), value.rbegin());
+}
+
```
vslavik, line 704: You already have `wxString::EndsWith()` at your disposal.

src/ui.cpp
```diff
@@ -696,8 +696,21 @@ void UpdateDialog::OnRunInstaller(wxCommandEvent&)
...
 bool UpdateDialog::RunInstaller()
 {
+    // Jsut let the app deal with it if it's a zip archive.
+    std::wstring _updateFile = m_updateFile;
+    if (ends_with(_updateFile, L".zip") && ApplicationController::_win_sparkle_zip_download_callback) {
+        ApplicationController::_win_sparkle_zip_download_callback(_updateFile);
+        return true;
```
vslavik, line 711: And what if it fails to apply?

vslavik (changes requested):
Sorry, but while I'd love to have proper ZIP support, I can't merge this in its current state; it is indeed very much "not very clean", even ignoring that broken second commit.

The fundamental issue is, of course, that this shouldn't be the hosting application's job; it should be done within WinSparkle. Although a callback is a possible interim solution, I'd rather do it right than have to carry compat baggage around forever. 

Also, please try to follow naming and formatting conventions of code you modify. Consistency matters.

translations/translations.rc
patrickkidd: I didn't know this would make it into the pull request. At any rate, visual studio 2019 complains about not finding the translation files when building vanilla win sparkle, even though the files exists. Seems like something wrong with the original solution/vc project.

.gitignore
patrickkidd: Again, this wasn't meant for your repo, sorry.

src/dll_api.cpp
patrickkidd: That is correct. I took a shortcut because I realized that adding a std::wstring dependency would break your C compatibility.

src/ui.cpp
patrickkidd: That is true. In terms of winsparkle concepts, do you think a useful alternative would be to allow the app to call a callback when the download payload is complete instead of the default WinSparkle behavior, regardless of file type? Makes sense if you want WinSparkle to keep the entire update workflow in WinSparkle, but how many use cases are unaccounted for? Maybe zip files containing a single aren't the only one? I am ignorant on this issue, but my case is one unaccounted for.

Also, there are enough classes and components in WinSparkle that it took too much energy to grok it all to determine the proper place for various functions. For example, I am not even clear that ```RunInstaller()``` is the correct place to fully handle a download payload like a zip file by: 1) move the running exe to the zip's temp folder (because you can't delete a running exe on windows), 2) unzip the new app exe from the downloaded zip file, and 3) move the new app exe in place, 4) launch the new app exe with ```CreateProcess(**)``` and exit the current process with ```ExitProcess(0)```. If you can suggest a cleaner way to handle this that doesn't break WinSparkle's design principles I can have a look at it.

patrickkidd:
Thanks for your thorough comments. My apologies for everything after the first commit. This is my first pull request and I did not know that they would make it into it. They were more for my own project.

At any rate, maybe if we can sort out the concepts here then I can submit a proper pull request that you can merge directly instead of just as a hypothetical suggestion for a new feature.

src/dll_api.cpp
vslavik: Make the callback take `const wchar_t*` argument (obtained with `wxString::wc_str()`) — it's read only and NULL-terminated, there's no risk there.

src/ui.cpp
vslavik: > That is true. In terms of winsparkle concepts, do you think a useful alternative would be to allow the app to call a callback when the download payload is complete instead of the default WinSparkle behavior, regardless of file type? 

Phrased like this, it makes sense to me — thanks for the insight! I was fixating on the ZIP case, but yes, a general callback mechanism to override download's handling is a good idea. 

> I am ignorant on this issue, but my case is one unaccounted for.

I don't feel the need to support *every* crazy case, but your one is reasonable, and it would be nice to support it properly. But until then, having the callback you propose is the pragmatic choice.

> I am not even clear that RunInstaller() is the correct place to fully handle a download payload like a zip file

It probably calls for putting it into a new class/file — as you say, a lot of logic is going to end up there.

> 1) move the running exe to the zip's temp folder (because you can't delete a running exe on windows),

It also needs to move all DLLs from the destination folder (or determine which ones are used by the current process) for the same reason, and it can't move them to the temp folder (permissions; inability to move to another drive). Which only supports your point that it is quite complicated for `RunInstaller()`.

> 4) launch the new app exe with CreateProcess(**) and exit the current process with ExitProcess(0). 

There's a callback for the latter in WinSparkle already (it may be more complicated and must e.g. ensure that files are saved and the user asked about it), so this too can get a bit more complicated.

> If you can suggest a cleaner way to handle this that doesn't break WinSparkle's design principles I can have a look at it.

I think it really needs separate class for all this (presumably specialized for different kinds - installers, ZIPs, delta updates eventually). 

But having an official, documented generic callback to apply the update, which would be used instead of the current launch code if present, is I think the best first step — it has the flexibility to handle unforeseen cases, is enough for your needs for now (if I understand it correctly) *and* you already mostly did it. 

TIA!

src/dll_api.cpp
patrickkidd: Roger that.

src/ui.cpp
patrickkidd: Roger.

src/ui.cpp
patrickkidd: Very good on all points. I'll start over to produce a viable pull request that simply adds a callback alternative in RunInstaller(). I say "alternative" because you do not want to execute a non-installer payload.

translations/translations.rc
patrickkidd: @vslavik Can you suggest a fix for this error? I could submit an issue on your repo but we are already here, so...

vslavik: > @vslavik Can you suggest a fix for this error?

That's the thing… *what* error? `master` builds [just fine](https://ci.appveyor.com/project/vslavik/winsparkle/builds/35923994).

patrickkidd: Visual studio complains about not finding the translation files listed in the resource file even though they exist. That happened in the zip source download as well as a clone from master.

vslavik: Again, see the link I provided for evidence to the contrary. If you believe there is genuine issue (on master, not on anything older), please report it, including all relevant details (including saying what you *did*), but I rather suspect from the overall tone of ad41ca4 that you opened the wrong (and converted) project file instead of using one directly usable by VS2019. There's only one project file on the `master` now, so that's a mistake no longer possible to make.

patrickkidd: Got it. I'll have a look at it again as I work on a pull request and report an issue on your repo if I can reproduce it.

PR #229, opened 2021-04-30
Add option to auto install after download when using win_sparkle_check_update_with_ui()

nickdademo:
When `win_sparkle_check_update_with_ui()` is used to allow the user to manually check for an update (i.e. via a menu item), the flow is the following (assuming an update is available):

1) Click menu item to check for updates
2) WinSparkle dialog shows that a new version is available
3) User clicks _"Install update"_

image

4) Update starts downloading
5) After download is complete, the user must then again click _"Install update"_

image

This PR allows the additional (and maybe redundant?) confirmation at Step (5) to be skipped and the update to be automatically installed after downloading, via an arg: `win_sparkle_check_update_with_ui(1);`

This is achieved in code by using two flags for the auto logic: one each for download and install

### DIFF
--- a/examples/psdk/app_psdk.c
+++ b/examples/psdk/app_psdk.c
@@ -66,7 +66,7 @@ void OnCheckForUpdates(HWND hwnd,
            always shows visible UI. You only need to call it if you have
            a "Check for updates" menu item, otherwise win_sparkle_init()
            checks for updates automatically. */
-        win_sparkle_check_update_with_ui();
+        win_sparkle_check_update_with_ui(0);
     }
 }
 
--- a/include/winsparkle.h
+++ b/include/winsparkle.h
@@ -553,6 +553,9 @@ WIN_SPARKLE_API void __cdecl win_sparkle_set_user_run_installer_callback(win_spa
     the user is told so. If there is an update, the usual "update available"
     window is shown.
 
+    Pass a non-zero value for @a auto_install to skip the "install"
+    confirmation after download and automatically install.
+
     This function returns immediately.
 
     @note Because this function is intended for manual, user-initiated checks
--- a/include/winsparkle.h
+++ b/include/winsparkle.h
@@ -561,7 +561,7 @@ WIN_SPARKLE_API void __cdecl win_sparkle_set_user_run_installer_callback(win_spa
 
     @see win_sparkle_check_update_without_ui()
  */
-WIN_SPARKLE_API void __cdecl win_sparkle_check_update_with_ui();
+WIN_SPARKLE_API void __cdecl win_sparkle_check_update_with_ui(int auto_install);
 
 /**
     Checks if an update is available, showing progress UI to the user and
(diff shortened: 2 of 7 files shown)


vslavik:
I don't think it makes much sense to have this configurable — if the current UI is bad, then it should be improved unconditionally.

The flow mirrors that of Sparkle for Mac, but perhaps with bad labels:

1. Release notes are shown, with *Install update* button.
2. After clicking it, the update is downloaded, with a *Cancel* button shown. This can take a while.
3. After the download completes, Sparkle shows *Install and relaunch* button (but WinSparkle repeats "Install update").

The reason for explicit confirmation in step 3 is that step 2 may be lengthy and unpredictable. You don't want the app to shut down under the user's hands a minute or two later, when they started working on something.

If I understand what you dislike correctly, then the "proper" solution is to add support for downloading updates in the background *before* the prompt with release notes is even shown (Sparkle for Mac supports this too). Then there would be a single *Install update* action, which would be instantaneous.

nickdademo:
> The reason for explicit confirmation in step 3 is that step 2 may be lengthy and unpredictable. You don't want the app to shut down under the user's hands a minute or two later, when they started working on something.

Definitely! Though, aren't we already doing this if we call `win_sparkle_check_update_with_ui_and_install()`? (lengthy download followed by an install with no confirmation - which could occur while the user returns to working with the app)

> If I understand what you dislike correctly, then the "proper" solution is to add support for downloading updates in the background before the prompt with release notes is even shown (Sparkle for Mac supports this too). Then there would be a single Install update action, which would be instantaneous.

Agreed - that would be a lot nicer, or alternatively: download starts when the dialog is opened (showing progress) and the "Install update" button is only enabled after the download is complete.

vslavik:
> Definitely! Though, aren't we already doing this if we call win_sparkle_check_update_with_ui_and_install()? (lengthy download followed by an install with no confirmation - which could occur while the user returns to working with the app)

Well, yes, but the context is a bit different, see https://github.com/vslavik/winsparkle/pull/77 — it's essentially meant for forced updates at launch.

Which part of the UI do you take issue with? That the button is labeled the same, or that an extra click is required? Renaming the first "Install update" to "Download update" probably makes sense to address the former...

nickdademo:
> 
> 
> > Definitely! Though, aren't we already doing this if we call win_sparkle_check_update_with_ui_and_install()? (lengthy download followed by an install with no confirmation - which could occur while the user returns to working with the app)
> 
> Well, yes, but the context is a bit different, see #77 — it's essentially meant for forced updates at launch.
> 
> Which part of the UI do you take issue with? That the button is labeled the same, or that an extra click is required? Renaming the first "Install update" to "Download update" probably makes sense to address the former...

The feedback from several customers has been something like this: _"We already made the decision to install the update when we first clicked 'Install update' - why do we have to click it again after the download?"_

I understand this feedback. So perhaps just renaming the first button to _"Download update"_ might be enough for now.

Though, I also think that auto-installing after the download is not so unexpected to the user as he/she had just clicked "Install update" a few seconds / minutes prior - this is the motivation for this PR.

Anyway, will close this PR for now - I'll investigate downloading the update earlier and making this a single-click workflow when I get some time.

Mapashito1:
If download is only a few seconds, I think it is interesting feature!! Users always complain about the second click

SyarifFakhri:
What's the plan with this PR? It's been open for 4 years. Will it get merged?

PR #230, opened 2021-05-06
Fix broken cmake support for MinGW32

daschuer:
This fixes building with mingw32 8.1 

Unfortunately wx fails with g++ 8.1 due to a wrong cast. 
This was fixed here: 
https://github.com/wxWidgets/wxWidgets/commit/424f64f27 
I can build after using the patch.


vslavik:
Sorry, I don't understand why does the commit titled "Update CMake support for MinGW32 8.1" do many things, none of which seem to be related to a particular MinGW version?

It's unclear to me why you add a generated config header to the repository when expat doesn't need it to be built on Windows (see: MSVS build projects).

daschuer:
The issue is that the cmake build in master are completely broken.

I have fist started ein MinGW 6.3, but the shipped header files do not contain all definitions required. The update to the recent MinGW 8.1 fixes the issue.

I did not notice that expat is not required. I just make it work, because it was part of the original cmake files.

When is expat required?

What needs to be done here, to put that int a mergable state?

vslavik:
> The issue is that the cmake build in master are completely broken.

Then that's what the commit message should say...

> I did not notice that expat is not required.

It is required. *Building it in this crazy-complicated way* is not. Again: see the VC++ build, which does not require a generated header you added. Ergo, evidently you don't need it either. Don't add it.

> What needs to be done here, to put that int a mergable state?

See above.

daschuer:
Done. 

Are the commit messages OK now? 

Did you consider how to proceed with the pending wx build error?

daschuer:
AppVeyor succeeds now

vslavik:
Thanks!

PR #239, opened 2022-01-16
Add preliminary EdDSA validation support

### LINKED ISSUE #187: Support EdDSA signatures
(Mac) Sparkle now implements more secure signatures using EdDSA (ed25519).

See [Sparkle 1.21.0 release](https://github.com/sparkle-project/Sparkle/releases/tag/1.21.0) and the bug that led to it — https://github.com/sparkle-project/Sparkle/issues/1037.

This doesn't appear to be documented anywhere except [the example](https://github.com/sparkle-project/Sparkle/blob/master/Resources/SampleAppcast.xml), but the signature is stored in the appcast as base64-encoded value in `sparkle::edSignature` property.

On our end, we discussed this in #157 particularly in regard to DSA being difficult on Windows. Sparkle uses a small portable library for it from https://github.com/orlp/ed25519 — we could use that too or we could use win32 API if possible. In any case, it's probably worth considering doing OpenSSL-less builds too, then.

/CCing @Youw as this may be of interest to you, and as I'd definitely appreciate your view on the ease (or not) of implementing ed25519 natively with win32 crypto API.

ntadej:
Add preliminary EdDSA validation support (partially closes #187 as signing is still missing). I tested with valid and invalid signature and it seems to work. Note that more testing is probably needed.

Some comments:
 - Uses https://github.com/orlp/ed25519.git like the primary Sparkle project, causes minimal size increase of the final dll.
 - Public key verfication is not there yet, I guess it is enough to just do the Base64 decoding + size check?
 - Not sure I updated bakefile/vcxproj correctly. I get some errors, but it everything looks fine. The only thing is that the format changed a bit (I guess I'm using a different version of bakefile as the current files were generated with).
 - Do we need that complicated file reading? I admit I took the first snippet from StackOverflow on how to read full files on windows.
 - RC integration not tested (I'm not primarily a Windows developer, I'm just trying to port one of my apps to Windows).

Tagging @kornelski and @vslavik.


purejava:
@ntadej, @Youw, do you see any chance to proceed with this PR?
Thanks.

purejava:
The reason why I am asking is, that I have created Java bindings for WinSparkle and we want to use these to update Cryptomator on Windows, that securely stores vaults in cloud services.

ntadej:
I am planning to, but Windows is currently at the bottom of priority list for my project.

geraldcombs:
I think I've fixed the issues identified by @Youw in https://github.com/geraldcombs/winsparkle/tree/eddsa-updates. The only outstanding issue is that SignatureVerifier::VerifyEdDSAPubKey is still a placeholder. I think the proper fix would be to add a corresponding function to the ed25519 public API.

@ntadej if my changes look OK to you, feel free to grab them and add them to this PR (and assign yourself authorship if you'd like). Otherwise I'd be happy to open a separate PR.

vslavik:
@geraldcombs Thanks!

> The only outstanding issue is that SignatureVerifier::VerifyEdDSAPubKey is still a placeholder. I think the proper fix would be to add a corresponding function to the ed25519 public API.

Let's make this simpler: `VerifyEdDSAPubKey` has little usefulness and the verification is done at signature checking time anyway, so we can either just remove this part or do what was initially suggested:

> Public key verfication is not there yet, I guess it is enough to just do the Base64 decoding + size check?

vslavik:
> Otherwise I'd be happy to open a separate PR.

Oh, sorry, forgot to say: please do.

ntadej:
Yes, sorry for lack of activity here. I also propose you open a separate PR so it will be easier to iterate for you.

geraldcombs:
> Oh, sorry, forgot to say: please do.

Done in #254.

vslavik:
→  #254

PR #244, opened 2022-04-28
Added limited support to the sparkle:tag criticalupdate

edwinclement08:
Updated the xml parser with a simple checker for whether criticalupdate
tag is present at present, it only detects the tag in this form, though
it can be extended to support the way sparkle 2 does with a bit of
effort.

```xml
 <sparkle:tags>
     <sparkle:criticalupdate/>
 </sparkle:tags>
```
p.s. this does not parse the specific version string embedded in the
criticalupdate tag. it also doesn't deal with the case where
criticalupdate tag appears outside of sparkle:tags

On detecting this particular tag in the appcast.xml, the installer
wizard will make the following changes to *stateupdateavailable*
- Hide the skip & remind me later buttons.
- Hide the close button on the title bar of the dialog box


3rdparty/expat.vcxproj
```diff
@@ -37,6 +37,7 @@
     <PlatformToolset Condition="'$(VisualStudioVersion)'=='15'">v141_xp</PlatformToolset>
     <PlatformToolset Condition="'$(VisualStudioVersion)'=='16'">v142</PlatformToolset>
     <XPDeprecationWarning>false</XPDeprecationWarning>
+    <PlatformToolset>v143</PlatformToolset>
```
Youw, line 40: I'm pretty sure it has to be:
`<PlatformToolset Condition="'$(VisualStudioVersion)'=='17'">v143</PlatformToolset>`
here and everywhere below

src/ui.cpp
```diff
@@ -534,23 +538,31 @@ UpdateDialog::UpdateDialog()
...
+    m_RemindLaterButtonSizer->Add(
                             new wxButton(this, ID_REMIND_LATER, _("Remind me later")),
-                            wxSizerFlags().Border(wxRIGHT, PX(10))
+                            wxSizerFlags()
                           );
+
+    m_buttonSizer->Add(m_SkipVersionButtonSizer, wxSizerFlags(0));
+    m_buttonSizer->AddStretchSpacer(1);
+    m_buttonSizer->Add(m_RemindLaterButtonSizer, wxSizerFlags(0));
+
```
Youw, line 559: spaces vs tabs

src/ui.cpp
```diff
@@ -895,6 +911,16 @@ void UpdateDialog::StateUpdateAvailable(const Appcast& info, bool installAutomat
...
         SHOW(m_updateButtonsSizer);
+        if (m_criticalUpdate) {
+            HIDE(m_SkipVersionButtonSizer);
+            HIDE(m_RemindLaterButtonSizer);
+            EnableCloseButton(false);
+        }
+        else {
+            SHOW(m_SkipVersionButtonSizer);
+            SHOW(m_RemindLaterButtonSizer);
+			EnableCloseButton(true);
```
Youw, line 922:
```suggestion
            EnableCloseButton(true);
```

edwinclement08:
Sorry about this, its not ready for review. Made a PR by mistake. Will clean it up and reopen

3rdparty/expat.vcxproj
edwinclement08: Updated that everywhere

src/ui.cpp
edwinclement08: Fixed this

src/ui.cpp
edwinclement08: Fixed this

edwinclement08:
Re-opening this, fixed code according to your comments, @Youw.
Also created this wiki entry as docs

src/appcast.cpp
```diff
@@ -113,6 +119,7 @@ bool is_windows_version_acceptable(const Appcast &item)
 void XMLCALL OnStartElement(void *data, const char *name, const char **attrs)
 {
     ContextData& ctxt = *static_cast<ContextData*>(data);
+    const size_t size = ctxt.items.size();
```
Youw, line 122: with `/W4` this will cause `C4456: declaration of 'size' hides previous local declaration` at line 170

edwinclement08: Removed the redundant declaration of size

src/ui.cpp
```diff
@@ -895,6 +911,16 @@ void UpdateDialog::StateUpdateAvailable(const Appcast& info, bool installAutomat
         HIDE(m_closeButtonSizer);
         HIDE(m_runInstallerButtonSizer);
         SHOW(m_updateButtonsSizer);
+        if (m_criticalUpdate) {
+            HIDE(m_SkipVersionButtonSizer);
+            HIDE(m_RemindLaterButtonSizer);
+            EnableCloseButton(false);
```
Youw, line 917: > EnableCloseButton(false);

Is it consistent with `(Mac)Sparkle` behavior?

Youw (commented):
No more comments from me.

src/ui.cpp
vslavik: I think it isn't - Sparkle [hides skip and later buttons](https://github.com/sparkle-project/Sparkle/blob/180f8975cfc18ef9816f69f963c7463f2f0425b4/Sparkle/SUUpdateAlert.m#L361). That makes sense - you want to prevent the user from ignoring the update and forgetting about it. But being able to close the UI entirely should IMHO be possible, because we don't know the circumstance - the user may e.g. be doing something urgent at the moment.

vslavik:
> Updated the xml parser with a simple checker for whether criticalupdate tag is present at present, it only detects the tag in this form, though it can be extended to support the way sparkle 2 does with a bit of effort.

What is the reason for doing it this way instead of supporting only Sparkle 2 way?

src/ui.cpp
edwinclement08: This is a requirement from my workplace. The install is still stoppable, after the download. The user can reject it in UAC prompt, or the later installer itself

edwinclement08:
@vslavik , honestly speaking, that was just time constraint and not a very good handle on writing C code.

edwinclement08:
Would be grateful for the help, but I would like it if this is merged first. We need it in our workplace and they might take a couple of weeks to get sorted out.

vslavik:
> Would be grateful for the help, but I would like it if this is merged first.

I'd rather not merge things only to immediately remove them. Looking at  the docs, it's actually *simpler* to extract `<sparkle:criticalUpdate>` from one level up than descending into now-obsolete `<sparkle:tags>`, so I see little reason to support the legacy syntax. 

> We need it in our workplace and the effort might take a couple of weeks to get sorted out.

Noted, will merge a.s.a.p.

Youw:
If your project is blocked by this change, you can always use your own fork instead of the original upstream.
Assepting half-solution is usually a bad idea and a nightmare for the developer to support.

src/ui.cpp
```diff
@@ -534,23 +538,31 @@ UpdateDialog::UpdateDialog()
...
-                            wxSizerFlags().Border(wxRIGHT, PX(20))
-                          );
-    m_updateButtonsSizer->AddStretchSpacer(1);
-    m_updateButtonsSizer->Add
-                          (
+                            wxSizerFlags()
+    );
+
+    m_RemindLaterButtonSizer = new wxBoxSizer(wxHORIZONTAL);
+    m_RemindLaterButtonSizer->Add(
```
vslavik, line 551: You don't need one sizer per control, you can use the `SHOW()` and `HIDE()` macros for individual controls too, not just sizers.

src/ui.cpp
```diff
@@ -534,23 +538,31 @@ UpdateDialog::UpdateDialog()
...
+    m_buttonSizer->AddStretchSpacer(1);
+    m_buttonSizer->Add(m_RemindLaterButtonSizer, wxSizerFlags(0));
+
     m_updateButtonsSizer->Add
                           (
                             m_installButton = new wxButton(this, ID_INSTALL, _("Install update")),
                             wxSizerFlags()
                           );
-    m_buttonSizer->Add(m_updateButtonsSizer, wxSizerFlags(1));
+    m_buttonSizer->Add(m_updateButtonsSizer, wxSizerFlags(0));
```
vslavik, line 565: I don't understand what's this for. If it is needed, then you can remove the entire ", wxSizerFlags(0)" part (that's the default) and the `AddStretchSpacer()` call above (there's nothing to stretch).

vslavik:
> @vslavik, If you are busy otherwise, can you please merge my PR?

I added some code review comments now. If you want this merged faster, please address them. They're minor details, though, the major issues preventing merging of this PR are the ones already discussed:

1. We should not use the legacy `<sparkle:tags>`, but should use the Sparkle 2 interface.
2. Disabling the close button is not OK.

vslavik:
VS2022 project support cherry-picked in b246977b.

vslavik:
> We should not use the legacy <sparkle:tags>, but should use the Sparkle 2 interface.

Oh, the code actually *already does* recognize both of the variants. 

I committed this as 7ca9b991, with changes from the review. I also simplified the appcast parsing parts (the code tracked descent into `<sparkle:tags>` but never actually used that information — turns out it supported the Sparkle 2 way already!). I fixed some typos. I radically simplified the UI part where much fewer changes were required. I also removed the hiding of close button — it really is too user hostile. The diff is much smaller now and may not look like it, but it wouldn't get there without your v1 of it.

Thanks for the code and thanks @Youw for your review.

PR #250, opened 2022-11-29
chore(ui): Catch fatal exceptions when showing release notes

jaredburkeen:
We've been getting several BugSplat crash reports related to exceptions  when showing the release notes. This is a pretty simple fix to catch those exceptions so that we don't crash.

### DIFF
--- a/3rdparty/expat.vcxproj
+++ b/3rdparty/expat.vcxproj
@@ -36,6 +36,7 @@
     <PlatformToolset Condition="'$(VisualStudioVersion)'=='14'">v140_xp</PlatformToolset>
     <PlatformToolset Condition="'$(VisualStudioVersion)'=='15'">v141_xp</PlatformToolset>
     <PlatformToolset Condition="'$(VisualStudioVersion)'=='16'">v142</PlatformToolset>
+    <PlatformToolset Condition="'$(VisualStudioVersion)'=='17'">v143</PlatformToolset>
     <XPDeprecationWarning>false</XPDeprecationWarning>
   </PropertyGroup>
   <PropertyGroup Label="Configuration" Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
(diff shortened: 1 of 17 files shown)


vslavik:
Why did you immediately close the PR?

jaredburkeen:
Sorry, this PR was intended to be based against our fork of the WinSparkle repo. We made several customizations so this PR doesn't really make sense to be merged to the original repo.
The only change that we actually made was to put a try-catch around `ShowReleaseNotes()`.
We did this because we've received several reports of customers running into a crash that appears to happen when that function is called.

Below is the call stack for the crash. The actual crash happens within the wxWidgets library, but we've been unable to identify a root cause.
That's why we added the try-catch. Also, we've been unable to reproduce the crash in our own internal testing.
Curious if you have run into this issue before or if there is a better approach to resolve the issue?
Screen Shot 2022-11-29 at 12 55 55 PM

vslavik:
@jaredburkeen Could you share the minidump file or full backtrace?

Normally, wxWidgets should have a catch-all within its even loop and this shouldn't be necessary. Full backtrace may provide insight into why it doesn't trigger.

jaredburkeen:
@vslavik Github won't allow me to share the dump file. I've emailed it you at vaclav@slavik.io.
Thank you for taking a look at it.

vslavik:
> @vslavik Github won't allow me to share the dump file. I've emailed it you at [vaclav@slavik.io](mailto:vaclav@slavik.io).

Yep. Sorry for not responding yet, but I'm aware and will look into it.

vslavik:
Sorry, I messed up and forgot I need the binaries to interpret minidump - did you use the "official" ones or given the existence of your fork, a custom build?  If the latter, could you get full backtrace from the minidump, as I won't be able to?

vslavik:
Browser handling changed significantly in 8f0490a4c5eb81df4765a76fe025bdc8ad8e9963, so I'm closing this as it is most likely fixed by that.  Please create a new issue or reopen if you see such things in WinSparkle >= 0.8.0

PR #254, opened 2023-05-08
EdDSA validation support

geraldcombs:
Build on @ntadej's work in #239 to add support for EdDSA validation. This should address the issues raised in that PR.


abemedia:
Is this still being worked on? Would be great to get ed25519 support in winsparkle!

ArchangelX360:
Hello, is this still be worked on? That would be awesome indeed to have parity between Sparkle and WinSparkle on signatures!

PR #265, opened 2023-12-27
Support LLVM MinGW with CMake.

tmiw:
This PR updates the CMake files as follows:

1. Updates the list of wxWidgets files to reflect updates to it since the CMake files were last touched.
2. Includes CMake logic from wxWidgets to pull in WebView2 and include its .lib/.dll files. 
3. For [LLVM MinGW](https://github.com/mstorsjo/llvm-mingw) only, uses a slightly modified `wx/setup.h` to require WebView2Loader.dll due to MinGW missing Microsoft's buffercheck code.
4. Minor fixes to allow WinSparkle to be built on Linux instead of Windows via cross-compiling.
5. Fixes a minor compiler error in the main WinSparkle code due to a missing `#include`.

Context: I'm one of the maintainers of the [FreeDV](https://freedv.org/) project and am looking to include an auto-update mechanism into our codebase (which uses CMake for configuration and LLVM MinGW to build the Windows binaries).

Let me know if you have any questions or need me to modify anything here.

### DIFF
--- a/3rdparty/WebView2/EventToken.h
+++ b/3rdparty/WebView2/EventToken.h
@@ -1,0 +1,69 @@
+
+
+/* this ALWAYS GENERATED file contains the definitions for the interfaces */
+
+
+ /* File created by MIDL compiler version 8.01.0622 */
+/* @@MIDL_FILE_HEADING(  ) */
+
+
+
+/* verify that the <rpcndr.h> version is high enough to compile this file*/
+#ifndef __REQUIRED_RPCNDR_H_VERSION__
+#define __REQUIRED_RPCNDR_H_VERSION__ 500
+#endif
+
+/* verify that the <rpcsal.h> version is high enough to compile this file*/
+#ifndef __REQUIRED_RPCSAL_H_VERSION__
+#define __REQUIRED_RPCSAL_H_VERSION__ 100
+#endif
+
+#include "rpc.h"
+#include "rpcndr.h"
+
+#ifndef __RPCNDR_H_VERSION__
+#error this stub requires an updated version of <rpcndr.h>
+#endif /* __RPCNDR_H_VERSION__ */
+
+
+#ifndef __eventtoken_h__
+#define __eventtoken_h__
+
+#if defined(_MSC_VER) && (_MSC_VER >= 1020)
+#pragma once
+#endif
+
+/* Forward Declarations */ 
+
+#ifdef __cplusplus
+extern "C"{
+#endif 
+
+
+/* interface __MIDL_itf_eventtoken_0000_0000 */
+/* [local] */ 
+
+//  Microsoft Windows
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+#pragma once
+typedef struct EventRegistrationToken
+    {
+    __int64 value;
+    } 	EventRegistrationToken;
+
+
+
+extern RPC_IF_HANDLE __MIDL_itf_eventtoken_0000_0000_v0_0_c_ifspec;
+extern RPC_IF_HANDLE __MIDL_itf_eventtoken_0000_0000_v0_0_s_ifspec;
+
+/* Additional Prototypes for ALL interfaces */
+
+/* end of Additional Prototypes */
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
+
+
--- a/3rdparty/wxWidgets_setup_h/clang/wx/setup.h
+++ b/3rdparty/wxWidgets_setup_h/clang/wx/setup.h
@@ -1,0 +1,1831 @@
+/////////////////////////////////////////////////////////////////////////////
+// Name:        wx/msw/setup.h
+// Purpose:     Configuration for the library
+// Author:      Julian Smart
+// Modified by:
+// Created:     01/02/97
+// Copyright:   (c) Julian Smart
+// Licence:     wxWindows licence
+/////////////////////////////////////////////////////////////////////////////
+
+#ifndef _WX_SETUP_H_
+#define _WX_SETUP_H_
+
+/* --- start common options --- */
+// ----------------------------------------------------------------------------
+// global settings
+// ----------------------------------------------------------------------------
+
+// define this to 0 when building wxBase library - this can also be done from
+// makefile/project file overriding the value here
+#ifndef wxUSE_GUI
+    #define wxUSE_GUI            1
+#endif // wxUSE_GUI
+
+// ----------------------------------------------------------------------------
+// compatibility settings
+// ----------------------------------------------------------------------------
+
+// This setting determines the compatibility with 2.8 API: set it to 0 to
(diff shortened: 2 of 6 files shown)


cmake/CMakeLists.txt
```diff
@@ -105,7 +105,12 @@ list(APPEND SOURCES ${CMAKE_CURRENT_BINARY_DIR}/common_res.o)
 
 add_library(${PROJECT_NAME} SHARED ${SOURCES} $<TARGET_OBJECTS:wxWidgets> $<TARGET_OBJECTS:expat> $<TARGET_OBJECTS:crypto>)
 
-target_link_libraries(${PROJECT_NAME} wininet version rpcrt4 comctl32 crypt32 wsock32 ws2_32 UxTheme Shlwapi)
+if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    # LLVM MinGW uses slightly different naming for UxTheme and Shlwapi.
+    target_link_libraries(${PROJECT_NAME} wininet version rpcrt4 comctl32 crypt32 wsock32 ws2_32 uxtheme shlwapi "${WEBVIEW2_LOADER_LIB}")
+else(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    target_link_libraries(${PROJECT_NAME} wininet version rpcrt4 comctl32 crypt32 wsock32 ws2_32 UxTheme Shlwapi)
+endif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
```
vslavik: Doesn't look like something that should matter on case-insensitive Windows filesystems?

cmake/wxWidgets/CMakeLists.txt
```diff
@@ -200,3 +251,36 @@ if(NOT PCHSupport_FOUND)
 endif()
 
 ADD_PRECOMPILED_HEADER(${PROJECT_NAME} ${SOURCE_DIR}/wxWidgets/include/wx/wxprec.h)
+
+set(WEBVIEW2_VERSION "1.0.705.50")
+set(WEBVIEW2_URL "https://www.nuget.org/api/v2/package/Microsoft.Web.WebView2/${WEBVIEW2_VERSION}")
+set(WEBVIEW2_SHA256 "6a34bb553e18cfac7297b4031f3eac2558e439f8d16a45945c22945ac404105d")
```
vslavik, line 261: This is going to be a PITA to maintain... I'd much prefer if it used the nuget tool or at least information from `packages.config`.

3rdparty/WebView2/EventToken.h
vslavik, line 1: Why is this file included?

3rdparty/wxWidgets_setup_h/clang/wx/setup.h
vslavik, line 1: Why is `setup.h` duplicated for Clang?

vslavik (changes requested):
I'd appreciate if this could follow the golden git rule of one commit per one self-contained change - right now some changes are split across two commits while the primary 79c5af5 commit does more than it says. In particular, wx/cmake and other compilation fixes are unrelated to Clang.

> For [LLVM MinGW](https://github.com/mstorsjo/llvm-mingw) only, uses a slightly modified wx/setup.h to require  WebView2Loader.dll due to MinGW missing Microsoft's buffercheck code.

This should be an `#ifdef` in the `setup.h` file instead of duplicated all of it.

cmake/CMakeLists.txt
Youw: It matters when cross-compiling using MinGW (or similar) from Linux host.

I don't remember what project was that, but the simplest solution we've used is to use the MinGW syntax (all lower-case) which works in all cases.

tmiw: I went ahead and simplified so we use lower-case for both Clang and not-Clang.

cmake/wxWidgets/CMakeLists.txt
tmiw: Added some simple regex-based parsing logic to grab the version from `packages.config`.

3rdparty/WebView2/EventToken.h
tmiw: WebView2 seems to need some definitions from this file. The file was originally retrieved from the https://github.com/jchv/webview2-in-mingw/ repo.

tmiw:
> I'd appreciate if this could follow the golden git rule of one commit per one self-contained change - right now some changes are split across two commits while the primary [79c5af5](https://github.com/vslavik/winsparkle/commit/79c5af5d869c6d4a19fbc1721e934359d471a461) commit does more than it says. In particular, wx/cmake and other compilation fixes are unrelated to Clang.

Understood, have tried to do so for commits after this message.

> This should be an `#ifdef` in the `setup.h` file instead of duplicated all of it.

Added `#if defined(__clang__)` in the existing setup.h instead of creating a new one.

3rdparty/WebView2/EventToken.h
vslavik: > WebView2 seems to need some definitions from this file.

Looking into this further,  I think you're duplicating a workaround that wxWidgets already includes, see https://github.com/wxWidgets/wxWidgets/commit/753253280005dde65a625493234193088941fa27

IOW, I *think* that changing the cmake to add `3rdparty/wxWidgets/include/wx/msw/wrl` to includedirs is all that should be needed instead of the new file?

tmiw: Looks like making that change did the trick, so I went ahead and removed EventToken.h.

vslavik:
Thanks! 

I'd like to add this to CI to avoid accidentally breaking it. Do I need to do anything special or will using MSYS2's `CLANG64` environment do?

tmiw:
> Thanks!
> 
> I'd like to add this to CI to avoid accidentally breaking it. Do I need to do anything special or will using MSYS2's `CLANG64` environment do?

I've never tried that environment, but I've basically been testing by doing something like the following:

```
$ cd ~
$ wget https://github.com/mstorsjo/llvm-mingw/releases/download/20231128/llvm-mingw-20231128-ucrt-ubuntu-20.04-x86_64.tar.xz
$ tar xf llvm-mingw-20231128-ucrt-ubuntu-20.04-x86_64.tar.xz
$ export PATH=~/llvm-mingw-20231128-ucrt-ubuntu-20.04-x86_64/bin:$PATH
$ cd winsparkle/cmake
$ mkdir build
$ cd build
$ cmake -DCMAKE_TOOLCHAIN_FILE=/path/to/llvm-mingw-toolchain.cmake ..
$ make
```

For llvm-mingw-toolchain.cmake, one of the files over at https://github.com/drowe67/freedv-gui/tree/master/cross-compile will work (depending on platform). Hope this helps!

vslavik:
Thanks for the pointers; I'll add this together with other CMake CI integration later, now that this is merged.

PR #266, opened 2023-12-27
Add logic to detect ARM updates from the appcast file.

tmiw:
This is a followup to #265 to enable WinSparkle to allow users to specifically provide ARM64 releases in the appcast file.

### DIFF
--- a/src/appcast.cpp
+++ b/src/appcast.cpp
@@ -213,11 +213,11 @@ bool is_suitable_windows_item(const Appcast &item)
 
     // Check suffix for matching bitness
 #ifdef _WIN64
-#ifdef defined(__AARCH64EL__) || defined(_M_ARM64)
+  #if defined(__AARCH64EL__) || defined(_M_ARM64)
     return item.Os.compare(OS_MARKER_LEN, std::string::npos, "-arm64") == 0;
-#else
+  #else
     return item.Os.compare(OS_MARKER_LEN, std::string::npos, "-x64") == 0;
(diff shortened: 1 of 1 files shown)


vslavik:
Thanks a lot!

I'd call it a follow-up to #50 instead (noting to get cross-referenced from there). Will also need to update https://github.com/vslavik/winsparkle/wiki/Appcast-Feeds accordingly after merging...

src/appcast.cpp
```diff
@@ -212,10 +213,14 @@ bool is_suitable_windows_item(const Appcast &item)
 
     // Check suffix for matching bitness
 #ifdef _WIN64
+#ifdef __AARCH64EL__ || _M_ARM64
```
vslavik, line 216: This is incorrect, it needs to be `#if defined(__AARCH64EL__) || defined(_M_ARM64)`.

Also, where does `__AARCH64EL__ ` come from? Can't find any reference to in relation to Windows...

tmiw: `__AARCH64EL__` is defined by LLVM MinGW:

```
parallels@ubuntu-linux-22-04-desktop:~/winsparkle/build$ aarch64-w64-mingw32-clang -E -dM - </dev/null | grep AARCH64
#define __AARCH64EL__ 1
#define __AARCH64_CMODEL_SMALL__ 1
parallels@ubuntu-linux-22-04-desktop:~/winsparkle/build$
```

Unfortunately it doesn't look like that version of MinGW defines `_M_ARM64`, so we can't simply check for that.

vslavik:
Thanks, merged!

Issue #271, opened 2024-05-10
Notify on WinSparkle update closed


BrunoBibboVoicemod:
Hi! I have a issue when closing the update dialog once download have finished. I give more context:

Our application do hide it's view if does found an update (we force the user to update to continue using the software). So if the user closes the update dialog when the package was already downloaded instead of click the "_install_" button, app will not be shown again as we don't receive any other event from WinSparkle.

When the updater finishes (closed by the user or installation starts) we would like run some checks after closing, but I don't see such behavior in this specific case above described.

We are using:
- WinSparkle V0.8.1
- Windows 11 Pro - Build 22631.3447

I took the liberty to check how was the "_quit_" behavior programmed and check how it would fit our needs. I found that in the `void UpdateDialog::OnClose(wxCloseEvent&)` method, at the end, there's an if preventing `ApplicationController::NotifyUpdateCancelled();` to be called if an update was already downloaded. I didn't want to mess with the current behavior of the library so I did just create a new event that will be executed always on WinSparkle update closed (yes, even when it closes successfully). This way I can suscribe to that behavior with a callback and do what I need to do without affecting the rest of usages.

I will submit a PR with the changes, what do you think?

Thanks in advance,
Bruno Bibbó

vslavik:
Duplicate of #127.

PRs are always welcome.

BrunoBibboVoicemod:
You're right, didn't saw the issue. PR submitted with the new callback method suggestion. Thanks!

Issue #127, opened 2017-02-02
win_sparkle_update_cancelled_callback not called when closing installation window


glaubitz:
Currently, the win_sparkle_update_cancelled_callback callback function is called when the user click "Skip Update", "Remind Me Later" or when the "Cancel" button is clicked when the update is being downloaded.

However, win_sparkle_update_cancelled_callback is not called when the user just closes the "Ready to install" dialog instead of clicking "Install updates" although this effectively cancels the update process as well.

vslavik:
Please submit a PR with the fix, then. As one of the *very* few users of this callback, you’re in the best position to do it and contribute back.

buwen-D:
@glaubitz Hello, I am also experiencing this problem now, and I need something to be done when the user closes the "Ready to install" dialog box, may I know how you fix it?

PR #272, opened 2024-05-10
Add update closed event

BrunoBibboVoicemod:
This is a solution candidate for #127 issue (duplicated #271) where WinSparkle library user can suscribe to a new event emitted when the updater closes (`win_sparkle_set_update_closed_callback`).

### DIFF
--- a/include/winsparkle.h
+++ b/include/winsparkle.h
@@ -459,6 +459,23 @@ typedef void (__cdecl *win_sparkle_update_cancelled_callback_t)();
 */
 WIN_SPARKLE_API void __cdecl win_sparkle_set_update_cancelled_callback(win_sparkle_update_cancelled_callback_t callback);
 
+/// Callback type for win_sparkle_update_closed_callback_t()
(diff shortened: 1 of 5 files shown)


vslavik:
Sorry, I still (hence the delay in response...) don't get why should this be a separate callback event and not just `win_sparkle_set_update_cancelled_callback` — it's effectively the same thing, isn't it?

It gets worse, though — I then noticed that there already is `win_sparkle_set_update_dismissed_callback`, contradicting the above opinion. And that one *is never actually called anywhere*.

So I *think*, but am open to arguments, that it should be separate event after all - simply because it was added to the API, however broken, a while back - *but* that this code should use `win_sparkle_set_update_dismissed_callback` instead of adding a new one.

vslavik:
I was wrong about the `dismissed` callback. The `dismissed` and `cancelled` callbacks are a hot mess, but adding yet another confusing one is not a solution, so I decided to close this PR and cleanup the existing callbacks instead.

PR #279, opened 2024-10-09
Dev/min autoupdate version

i-n-g-o:
Hello.

This PR adds the feature to use minimumAutoupdateVersion in appcast.xml.
It also fixes some other issues:
- handles multiple enclosure-tags by only accepting "windows" as operating system. Not specifying the os is still valid.
- IsValid also checks if Os is empty or starts with "windows"
- corrects the closing of item-tag

I tried to keep the original behaviour intact and committed each step for better clarity. Please let me know what you think.

### DIFF
--- a/src/appcast.cpp
+++ b/src/appcast.cpp
@@ -25,6 +25,8 @@
 
 #include "appcast.h"
 #include "error.h"
+#include "settings.h"
+#include "updatechecker.h"
 
 #include <expat.h>
 #include <vector>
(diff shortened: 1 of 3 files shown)


i-n-g-o:
Thanks for your reply and for checking this PR.

Erm, yes. Despite its clear naming the tag was somehow understood as a general minimumVersion for the update and is applied as such. Yes, this seems wrong - sorry for the noise.

Sure, I will provide individual PRs and try to be clearer with my explanations with proper formatting.

PR #280, opened 2024-10-17
Correct closing of NODE_ITEM

i-n-g-o:
When OnEndElement(...) is called ctxt.in_item is still > 0. Therefore checking if the closing node is NODE_ITEM needs to happen inside that conditional block to decrease ctxt.in_item and possibly stop parsing.

### DIFF
--- a/src/appcast.cpp
+++ b/src/appcast.cpp
@@ -273,12 +273,12 @@ void XMLCALL OnEndElement(void *data, const char *name)
         {
             ctxt.in_dsasignature--;
         }
-    }
-    else if (ctxt.in_channel && strcmp(name, NODE_ITEM) == 0)
-    {
-        ctxt.in_item--;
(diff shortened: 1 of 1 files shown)


vslavik:
I'm sorry, but even after several re-readings, I don't understand *why* you want to make the change. The commit message explains *what* the change is (something the code already answers), but not *why*.

Can you please explain what is handled incorrectly and provide an example feed content that demonstrates it? TIA!

i-n-g-o:
As far as I can see the code to preliminary stop parsing can never be called. It seemed not right so that is why I proposed this PR. Maybe I am missing something and the original code is perfectly fine - in that case this PR of course can be safely discarded.

vslavik:
No, you're right, I just couldn't understand what you were saying, because that bug was so absurd it didn't register... Sorry for being slow!

PR #281, opened 2024-10-17
Only overwrite with suitable enclosure data

i-n-g-o:
If just overwriting enclosure data multiple enclosure-tags may result in wrong update information. Consider an update for multiple platforms where the enclosure tag for windows is not the last tag.

This change only allows enclosure data to be overwritten if it is suitable for window. It leaves the current behaviour intact by accepting enclosure data if no Os is set.

### DIFF
--- a/src/appcast.cpp
+++ b/src/appcast.cpp
@@ -111,6 +111,37 @@ bool is_windows_version_acceptable(const Appcast &item)
         VER_SERVICEPACKMAJOR, dwlConditionMask) != FALSE;
 }
 
+
+/**
+ * Returns true if item os is exactly "windows"
+ *   or if item is "windows-arm64" on 64bit ARM
+ *   or if item is "windows-x64" on 64bit Intel/AMD
+ *   or if item is "windows-x86" on 32bit
+ *   and is above minimum version
+ */
+bool is_suitable_windows_item(const Appcast& item)
+{
+    if (!is_windows_version_acceptable(item))
+        return false;
+
+    if (item.Os == OS_MARKER)
+        return true;
+
+    if (item.Os.compare(0, OS_MARKER_LEN, OS_MARKER) != 0)
+        return false;
+
+    // Check suffix for matching bitness
+#ifdef _WIN64
+#if defined(__AARCH64EL__) || defined(_M_ARM64)
+    return item.Os.compare(OS_MARKER_LEN, std::string::npos, "-arm64") == 0;
+#else
+    return item.Os.compare(OS_MARKER_LEN, std::string::npos, "-x64") == 0;
+#endif // defined(__AARCH64EL__) || defined(_M_ARM64)
+#else
+    return item.Os.compare(OS_MARKER_LEN, std::string::npos, "-x86") == 0;
+#endif // _WIN64
+}
+
 void XMLCALL OnStartElement(void *data, const char *name, const char **attrs)
 {
     ContextData& ctxt = *static_cast<ContextData*>(data);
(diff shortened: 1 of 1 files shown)


vslavik:
This is same situation as the other PR: can you please provide example feed demonstrating the issue?

src/appcast.cpp
```diff
@@ -163,24 +194,40 @@ void XMLCALL OnStartElement(void *data, const char *name, const char **attrs)
...
+                // This might set the wrong operating system but accepts enclosures
+                // without Os information.
+                if (item.Os.empty() || is_suitable_windows_item(enclosure_item))
+                {
+                    item.Os = enclosure_item.Os;
+                    item.DownloadURL = enclosure_item.DownloadURL;
+                    item.Version = enclosure_item.Version;
+                    item.ShortVersionString = enclosure_item.ShortVersionString;
+                    item.DsaSignature = enclosure_item.DsaSignature;
+                    item.InstallerArguments = enclosure_item.InstallerArguments;
```
vslavik, line 230: This is error-prone and will *silently* break if any attribute is added to `Appcast`. If such copying is needed (it's not entirely clear to me why), it should be handled by having a copyable sub-object for the enclosure in the `Appcast` class.

i-n-g-o:
> This is same situation as the other PR: can you please provide example feed demonstrating the issue?

Sure, here are two appcast xmls showing the same issue with different data.

[appcast-win-arch.xml.txt](https://github.com/user-attachments/files/17509250/appcast-win-arch.xml.txt)
[appcast-win-mac.xml.txt](https://github.com/user-attachments/files/17509251/appcast-win-mac.xml.txt)

src/appcast.cpp
i-n-g-o: Hm, I would argue it is not more error prone than it was before: It would also silently break if more attributes are added in the original version.

It seemed handy to use Appcast but maybe individual strings or an extra structure for the attributes would make the code more readable.

The reason for the extra copy-step is to avoid directly overwriting data in Appcast without checking the data beforehand.

Initially I thought of introducing a structure for the enclosure data. But I could not find a clear documentation of the appcast.xml and it seems that it evolved over time and places for information changed. (e.g.: "version" can be a node and/or can be set in the enclosure tag?)
My conclusion was to not introduce any extra structure and use what is there to make minimal changes only.

vslavik:
Thanks for the examples and the PR!

While your patch itself is OK in that it corrects this issue, there were more (e.g. not ignoring items w/o a usable enclosure, not prioritizing `windows-$arch` over `windows` os value etc.) and so I felt like it's past time I rewrite the code in a more robust, followable way, and will merge PR #283 instead.

However, your comments, examples and, indeed, code from this PR, helped me a lot in understanding all the subtle bugs, and I appreciate your help a lot.

i-n-g-o:
Cool. That's great!

I will give it a try as soon as I can. Thanks for the fix!

vslavik:
> I will give it a try as soon as I can. 

Appreciated! I'm reasonably confident in the fix (it passes your examples), but the more eyes, the better.

i-n-g-o:
> Appreciated! I'm reasonably confident in the fix (it passes your examples), but the more eyes, the better.

Hi. I looked at the changes in https://github.com/vslavik/winsparkle/pull/283. This all looks very good to me.

Do you think this would already be worth a new tagged version (0.8.2)?

vslavik:
> Do you think this would already be worth a new tagged version (0.8.2)?

Yep: https://github.com/vslavik/winsparkle/releases/tag/v0.8.2

PR #282, opened 2024-10-17
Check Os in IsValid()

i-n-g-o:
If a non-windows Os is set the update item should not be valid. Consider an update item with an enclosure for a non-windows version.

This change additionally checks if Os is either empty or starts with OS_MARKER. It also moves the implementation into appcast.cpp so it has access to OS_MARKER_LEN and OS_MARKER.

### DIFF
--- a/src/appcast.cpp
+++ b/src/appcast.cpp
@@ -386,4 +386,15 @@ Appcast Appcast::Load(const std::string& xml)
     }
 }
 
+
+/**
+* Returns true if Version is not empty and
+* Os is empty or starts with OS_MARKER
+*/
+bool Appcast::IsValid() const
+{
+    return !Version.empty() &&
+        (Os.empty() || Os.compare(0, OS_MARKER_LEN, OS_MARKER) == 0);
+}
+
 } // namespace winsparkle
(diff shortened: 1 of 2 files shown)


vslavik:
This one I am sure about: it shouldn't be merged (sorry!). `IsValid()` is explicitly (see the *Returns true if the struct constains valid data* comment) about validity of the data, i.e. difference between uninitiated and filled-in struct, not about making judgement about the item's suitability under some runtime conditions.

i-n-g-o:
Would it then make sense to add another check about the runtime condition?

Consider this appcast:
```
<?xml version="1.0" encoding="UTF-8"?>
<rss version="2.0" xmlns:sparkle="http://www.andymatuschak.org/xml-namespaces/sparkle" xmlns:dc="http://purl.org/dc/elements/1.1/">
  <channel>
    <title>Test</title>
    <language>en</language>

    <item>
      <title>Version 1.0.2</title>
      <sparkle:version>1.0.2</sparkle:version>
      <enclosure sparkle:os="macos" url="url-to-update.dmg"/>
    </item>
  </channel>
</rss>
```

Currently this would be a valid update, would be offered and the dmg would be downloaded. Yes it might be misconfigured for windows, but I think it should not result in a valid update.

PR #287, opened 2025-04-22
VerifyEdDSAPubKey should take arg pubkey

APshenkin:
VerifyEdDSAPubKey should take argument value for verification

Othewise Settings:SetEdDSAPubKey doesn't verify it.

### DIFF
--- a/src/signatureverifier.cpp
+++ b/src/signatureverifier.cpp
@@ -298,7 +298,7 @@ void SignatureVerifier::VerifyDSAPubKeyPem(const std::string &pem)
 
 void SignatureVerifier::VerifyEdDSAPubKey(const std::string& pubkey_base64)
 {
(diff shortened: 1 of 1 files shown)


vslavik:
Thanks!

APshenkin:
@vslavik any chance to make release for this version that will contain a fix?

vslavik:
Sorry, I misunderstood the impact to be what you wrote, i.e. that `Settings:SetEdDSAPubKey` doesn't verify the key against programmer error.

I didn't realize that the *actual* impact is much more severe, i.e. *preventing the pubkey from being set at all* when using `win_sparkle_set_eddsa_public_key()` and not the `EdDSAPub` resource.

The former wouldn't warrant a dedicated release, the latter does.

vslavik:
→  [v0.9.1](https://github.com/vslavik/winsparkle/releases/tag/v0.9.1)

PR #290, opened 2026-02-13
Fix private key file arg in sign example command

mminer:
Minor documentation fix: I noticed that the example command for signing in the README is incorrect. It uses a `--file` argument, when it should be `--private-key-file`.


vslavik:
Thank you!

PR #294, opened 2026-05-06
Fix release notes when URL is redirected

### LINKED ISSUE #293: Redirected Release Notes URL opens in external browser
If a release notes URL returns HTTP 302 then WinSparkle launches it in an external browser and leaves the webview blank. WinSparkle used to handle a redirected URL as expected, but this new behavior appears to be a result of the 2023 change to handle clicked links: ec271892eaefd48edffd9ec2123fd9f9f09f5113. It looks like the `wxEVT_WEBVIEW_NAVIGATING` event fires on the HTTP 302 and is treated as a click.

We got bit by this when we upgraded from 0.6.0 to 0.9.0. We worked around it by updating our URL to avoid the redirect, but I thought you'd like to know. I don't know if `wxEVT_WEBVIEW_NAVIGATING` exposes the difference between a click and a redirect, but maybe `wxWebViewEvent::GetNavigationAction()` can be used to differentiate.

I suspect the older machine this was first observed on was using the IE backend for the webview. I didn't know about webview backends until after we'd updated our URL, so I don't know if the backend is relevant.

Related: wxWidgets changed this behavior in 3.1.2: https://github.com/wxWidgets/wxWidgets/blob/f57b17ba5f63db53dd1b465646ade105fee22302/docs/changes_32.txt#L824

Thank you for WinSparkle! =)

sluicebox:
This fixes https://github.com/vslavik/winsparkle/issues/293 where release notes are loaded in an external browser if the HTTP response is a redirect. This became an issue in ec271892eaefd48edffd9ec2123fd9f9f09f5113, when a `wxEVT_WEBVIEW_NAVIGATING` event handler was added to open clicked links in the browser. The problem is that a redirect generates an additional `wxEVT_WEBVIEW_NAVIGATING` event with the new URL, and that event was treated as a click.

`wxEVT_WEBVIEW_NAVIGATING` does not contain information to differentiate between an initial redirect and a clicked link. There is an action field, but it is not set by this event.

Instead, we can wait for the `wxEVT_WEBVIEW_LOADED` event, and treat any `wxEVT_WEBVIEW_NAVIGATING` event that occurs afterwards as a click. All of the initial `wxEVT_WEBVIEW_NAVIGATING` events occur before `wxEVT_WEBVIEW_LOADED`.

Tested:
- A release notes URL that redirects (HTTP 301 and 302)
- A release notes URL that does not redirect
- A HTML description string instead of a release notes URL

I left in the `url.starts_with("http")` check to be safe, but I think that only existed to ignore `wxEVT_WEBVIEW_NAVIGATING` when using an HTML description string instead of a release notes URL. If that's all that check was for, then it can be removed.

I only have access to an older Windows machine currently, so I've only been able to test this with the IE backend.

### DIFF
--- a/src/ui.cpp
+++ b/src/ui.cpp
@@ -482,6 +482,8 @@ private:
     bool m_errorOccurred;
     // whether window closure was updater-initiated (i.e. not caused by user)
     bool m_closeInitiatedByUpdater;
+    // whether the web browser has loaded the release notes
+    bool m_webBrowserLoaded;
 
     static const int RELNOTES_WIDTH = 460;
     static const int RELNOTES_HEIGHT = 200;
@@ -495,6 +497,7 @@ UpdateDialog::UpdateDialog()
     m_installAutomatically = false;
     m_errorOccurred = false;
     m_closeInitiatedByUpdater = false;
+    m_webBrowserLoaded = false;
 
     m_heading = new wxStaticText(this, wxID_ANY, "");
     SetHeadingFont(m_heading);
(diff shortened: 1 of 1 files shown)


vslavik:
Sorry for the delay; I was traveling.

Thanks a lot for your analysis and the patch; looks good to me!

> I only have access to an older Windows machine currently, so I've only been able to test this with the IE backend.

That’s actually great: I’m the other way around: can test with Edge easily, but less so with IE.

PR #296, opened 2026-06-24
Adds support for silent, in-app installation of MSIX / MSIXBundle updates.

APshenkin:
Adds support for silent, in-app installation of MSIX / MSIXBundle updates*

Today WinSparkle launches downloaded installers with `ShellExecuteEx`. That doesn't work for `.msix`/`.msixbundle` packages: the shell association for those is the **App Installer GUI**, so the user is bounced out into a separate Microsoft window and has to click through it manually.

This PR lets an app opt into a fully silent flow by marking the appcast enclosure with `sparkle:installerType="msix"`. WinSparkle then installs the package through the OS deployment API (`Windows.Management.Deployment.PackageManager.AddPackageAsync`), shows staging progress in its own dialog, and relaunches the app once the update is applied.

[MSIX](https://learn.microsoft.com/windows/msix/overview) is Microsoft's modern app packaging format for Windows — it gives any app a reliable clean install/uninstall, differential (block-level) updates, and a package identity that unlocks many Windows platform features. It's also the format the **Microsoft Store requires for published apps**, and it's increasingly used for direct, outside-the-Store distribution.

## Demo

**1. Silent MSIX update — happy path (download → progress → relaunch):**

https://github.com/user-attachments/assets/44fc73f6-2eef-47bf-9afe-157574302e14

**2. How it looks before the changes**

https://github.com/user-attachments/assets/205d4b18-54a6-4889-9fcf-e84a11c626ca

## How to use it

Mark the enclosure in your appcast with the new `sparkle:installerType` attribute:

```xml
<enclosure
    url="https://example.com/MyApp_1.2.3.0_x64.msixbundle"
    sparkle:version="1.2.3.0"
    sparkle:installerType="msix"
    length="12345678"
    type="application/octet-stream" />
```

Anything other than `msix` (including an absent attribute) keeps the existing `ShellExecuteEx` behavior, so this is fully backward compatible.

## How it works

- **Appcast** — parses the new `sparkle:installerType` enclosure attribute (`Appcast::Enclosure::InstallerType`).
- **Install path** — when the type is `msix`, `OnRunInstaller` calls `StartMsixInstall()` instead of launching the file. The install runs on a background thread; progress and completion are marshaled back to the UI thread via `UI::NotifyInstallProgress` / `UI::NotifyInstallFinished`.
- **UI** — new dialog states: `StateInstalling` (determinate progress bar), `InstallProgress` (0–100%), and `StateInstallFinished`. On success the app shuts down so Windows can swap the package files; on failure it shows the error and returns to the "ready to install" state so the user can retry.
- **Relaunch** — `RegisterApplicationRestart` is called before `AddPackageAsync(..., ForceApplicationShutdown)`, so Windows relaunches the app after the files are swapped during shutdown.

## Compatibility & fallback

The MSIX path must not break existing consumers on older Windows or non-MSVC builds:

- **Pre-Windows 10** — the WinRT runtime DLLs are **delay-loaded** (see `DelayLoadDLLs` in the vcxproj), and the code probes `api-ms-win-core-winrt-l1-1-0.dll` at runtime (`IsMsixDeploymentAvailable`) before touching any C++/WinRT code. So `WinSparkle.dll` still loads and runs normally on systems that predate MSIX; the feature simply reports as unsupported.
- **Non-MSVC / MinGW builds** — `msix.cpp` guards on `__has_include(<winrt/...>)` and compiles to a stub that reports "not supported in this build" when the Windows 10 SDK headers aren't present.
- `msix.cpp` is built with **C++17** (required by the C++/WinRT projection); the rest of the project is unaffected.

## Changes

| File | Change |
|------|--------|
| `src/msix.cpp`, `src/msix.h` | **New** — silent MSIX install via `PackageManager`, runtime availability probe, background worker, MinGW stub |
| `src/appcast.cpp`, `src/appcast.h` | Parse `sparkle:installerType` attribute |
| `src/ui.cpp`, `src/ui.h` | Installing/progress/finished states + thread-safe notifications |
| `WinSparkle.vcxproj` | Link `runtimeobject.lib`/`delayimp.lib`, delay-load WinRT DLLs, build `msix.cpp` as C++17 |
| `cmake/CMakeLists.txt` | Add `msix.cpp` to sources |
| `NEWS` | 0.9.4 changelog entry |

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
   build:
     name: Build DLL
-    runs-on: windows-latest
+    runs-on: windows-2022
     strategy:
       matrix:
         platform: [Win32, x64, ARM64]
--- a/src/msix.cpp
+++ b/src/msix.cpp
@@ -49,6 +49,7 @@
 #include <shlwapi.h>
 
 #include "utils.h"
+#include "threads.h"
 
 #include <winrt/Windows.Foundation.h>
 // Required for the IIterable<Uri> parameter of AddPackageAsync(); without it the
@@ -58,7 +59,6 @@
 
 #include <string>
 #include <stdexcept>
-#include <thread>
 
 // The WinRT runtime imports are delay-loaded (see the /DELAYLOAD entries in the
 // project file) so that WinSparkle.dll still loads on Windows versions that
@@ -166,14 +166,49 @@ void InstallWorker(std::wstring packagePath)
         winrt::uninit_apartment();
 }
 
+
+// This deliberately does not run on the UI thread: that thread is a
+// single-threaded COM apartment (wxWidgets initializes it via OleInitialize), and
+// the blocking DeploymentOperation::get() call in InstallWorker() is illegal there
+// -- C++/WinRT asserts on it in debug builds and it would stall the message pump
+// (deadlock-prone) in release builds. This worker initializes a multi-threaded
+// apartment instead, where the blocking wait is legal and safe.
+//
+// Fire-and-forget: the thread self-destructs once Run() returns
+// (IsJoinable() == false), matching the previous detached-thread behavior.
+class MsixInstallThread : public Thread
+{
+public:
+    explicit MsixInstallThread(const std::wstring& packagePath)
+        : Thread("WinSparkle MSIX installer"), m_packagePath(packagePath)
+    {
+    }
+
+protected:
+    void Run() override
+    {
+        // No initialization that Start() must wait for, so signal readiness
+        // immediately; the deployment below can take a long time and must not
+        // block the (UI) thread that called Start().
+        SignalReady();
+        InstallWorker(m_packagePath);
+    }
+
+    bool IsJoinable() const override { return false; }
+
+private:
+    std::wstring m_packagePath;
+};
+
 } // anonymous namespace
 
 
 void StartMsixInstall(const std::wstring& packagePath)
 {
-    // The deployment runs off the UI thread; progress and completion are reported
-    // back asynchronously via UI::NotifyInstall*().
-    std::thread(InstallWorker, packagePath).detach();
+    // The deployment runs off the UI thread (see MsixInstallThread); progress and
+    // completion are reported back asynchronously via UI::NotifyInstall*(). The
+    // thread is fire-and-forget and deletes itself when finished.
+    (new MsixInstallThread(packagePath))->Start();
 }
 
 } // namespace winsparkle
--- a/src/msix.cpp
+++ b/src/msix.cpp
@@ -60,14 +60,6 @@
 #include <string>
 #include <stdexcept>
 
-// The WinRT runtime imports are delay-loaded (see the /DELAYLOAD entries in the
-// project file) so that WinSparkle.dll still loads on Windows versions that
-// predate MSIX. runtimeobject.lib provides the Ro*/Windows*String imports that
-// the C++/WinRT projection calls into; if a future SDK requires windowsapp.lib
-// instead, adjust the project's linker inputs accordingly.
-#pragma comment(lib, "runtimeobject.lib")
-#pragma comment(lib, "shlwapi.lib")
-
 namespace winsparkle
 {
 
--- a/src/msix.cpp
+++ b/src/msix.cpp
@@ -1,0 +1,190 @@
+/*
+ *  This file is part of WinSparkle (https://winsparkle.org)
+ *
+ *  Copyright (C) 2009-2026 Vaclav Slavik
+ *
+ *  Permission is hereby granted, free of charge, to any person obtaining a
+ *  copy of this software and associated documentation files (the "Software"),
+ *  to deal in the Software without restriction, including without limitation
+ *  the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ *  and/or sell copies of the Software, and to permit persons to whom the
+ *  Software is furnished to do so, subject to the following conditions:
+ *
+ *  The above copyright notice and this permission notice shall be included in
+ *  all copies or substantial portions of the Software.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ *  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ *  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ *  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ *  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ *  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ *  DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+#include "msix.h"
+#include "ui.h"
+#include "error.h"
+
+// MSIX silent installation relies on the C++/WinRT projection of
+// Windows.Management.Deployment, which ships with the Windows 10 SDK. When those
+// headers aren't available (e.g. a MinGW build), this file compiles to a stub
+// that simply reports the feature as unsupported.
+#if defined(__has_include)
+    #if __has_include(<winrt/Windows.Management.Deployment.h>)
+        #define WINSPARKLE_HAVE_MSIX 1
+    #endif
+#endif
+
+#ifdef WINSPARKLE_HAVE_MSIX
+
+#ifndef WIN32_LEAN_AND_MEAN
+    #define WIN32_LEAN_AND_MEAN
+#endif
+#ifndef NOMINMAX
+    #define NOMINMAX
+#endif
+#include <windows.h>
+#include <shlwapi.h>
+
+#include "utils.h"
+
+#include <winrt/Windows.Foundation.h>
+#include <winrt/Windows.Management.Deployment.h>
+
+#include <string>
+#include <stdexcept>
+#include <thread>
+
+// The WinRT runtime imports are delay-loaded (see the /DELAYLOAD entries in the
+// project file) so that WinSparkle.dll still loads on Windows versions that
+// predate MSIX. runtimeobject.lib provides the Ro*/Windows*String imports that
+// the C++/WinRT projection calls into; if a future SDK requires windowsapp.lib
+// instead, adjust the project's linker inputs accordingly.
+#pragma comment(lib, "runtimeobject.lib")
+#pragma comment(lib, "shlwapi.lib")
+
+namespace winsparkle
+{
+
+namespace
+{
+
+// Probe whether the WinRT deployment API is present before touching any C++/WinRT
+// code. The API set below is absent on pre-Win10 systems; calling into a WinRT
+// function there would trigger a delay-load failure (a structured exception that
+// /EHsc does not catch). Probing first keeps us safe and lets us report cleanly.
+bool IsMsixDeploymentAvailable()
+{
+    HMODULE mod = ::LoadLibraryExW(
+        L"api-ms-win-core-winrt-l1-1-0.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
+    if (!mod)
+        return false;
+    ::FreeLibrary(mod);
+    return true;
+}
+
+// Turn a local filesystem path into a properly escaped file:// URI, which is what
+// PackageManager.AddPackageAsync() expects.
+std::wstring MakeFileUri(const std::wstring& path)
+{
+    wchar_t url[4096];
+    DWORD len = ARRAYSIZE(url);
+    HRESULT hr = ::UrlCreateFromPathW(path.c_str(), url, &len, 0);
+    if (FAILED(hr))
+        throw std::runtime_error("Cannot build file URI for MSIX package");
+    return std::wstring(url);
+}
+
+void InstallWorker(std::wstring packagePath)
+{
+    if (!IsMsixDeploymentAvailable())
+    {
+        LogError("MSIX deployment API is not available on this version of Windows.");
+        UI::NotifyInstallFinished(false, L"MSIX installation is not supported on this version of Windows.");
+        return;
+    }
+
+    bool apartmentInited = false;
+    try
+    {
+        winrt::init_apartment(winrt::apartment_type::multi_threaded);
+        apartmentInited = true;
+
+        // Register for automatic restart *before* applying the update, so that
+        // Windows relaunches the app after ForceApplicationShutdown swaps files.
+        ::RegisterApplicationRestart(nullptr, 0);
+
+        namespace deploy = winrt::Windows::Management::Deployment;
+
+        deploy::PackageManager manager;
+        auto op = manager.AddPackageAsync(
+            winrt::Windows::Foundation::Uri(MakeFileUri(packagePath)),
+            nullptr,
+            deploy::DeploymentOptions::ForceApplicationShutdown);
+
+        op.Progress([](auto const&, deploy::DeploymentProgress const& progress)
+        {
+            UI::NotifyInstallProgress(static_cast<int>(progress.percentage));
+        });
+
+        deploy::DeploymentResult result = op.get();
+
+        if (result.ExtendedErrorCode() < 0)
+        {
+            std::wstring err(result.ErrorText().c_str());
+            LogError("MSIX deployment failed: " + WideToAnsi(err));
+            UI::NotifyInstallFinished(false, err);
+        }
+        else
+        {
+            UI::NotifyInstallFinished(true, std::wstring());
+        }
+    }
+    catch (const winrt::hresult_error& ex)
+    {
+        std::wstring err(ex.message().c_str());
+        LogError("MSIX deployment error: " + WideToAnsi(err));
+        UI::NotifyInstallFinished(false, err);
+    }
+    catch (const std::exception& ex)
+    {
+        LogError(std::string("MSIX deployment error: ") + ex.what());
+        UI::NotifyInstallFinished(false, L"MSIX installation failed.");
+    }
+    catch (...)
+    {
+        LogError("MSIX deployment failed with an unknown error.");
+        UI::NotifyInstallFinished(false, L"MSIX installation failed.");
+    }
+
+    if (apartmentInited)
+        winrt::uninit_apartment();
+}
+
+} // anonymous namespace
+
+
+void StartMsixInstall(const std::wstring& packagePath)
+{
+    // The deployment runs off the UI thread; progress and completion are reported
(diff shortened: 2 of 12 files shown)


src/ui.cpp
```diff
@@ -683,6 +695,17 @@ void UpdateDialog::OnRunInstaller(wxCommandEvent&)
         return;
     }
 
+    // MSIX/MSIXBundle packages cannot be installed silently via ShellExecuteEx
+    // (the shell would launch the App Installer GUI). Install them through the OS
+    // deployment API instead, showing progress in this dialog. Completion is
+    // handled asynchronously in StateInstallFinished().
+    if ( wxString(m_installerType.c_str()).IsSameAs("msix", false) )
```
vslavik: WTF is this form of `operator==`? 😳

src/msix.cpp
```diff
@@ -0,0 +1,193 @@
...
+}
+
+} // anonymous namespace
+
+
+void StartMsixInstall(const std::wstring& packagePath)
+{
+    // The deployment runs off the UI thread; progress and completion are reported
+    // back asynchronously via UI::NotifyInstall*().
+    std::thread(InstallWorker, packagePath).detach();
```
vslavik: This codebase uses `wxThread`

src/ui.cpp
```diff
@@ -983,14 +1006,100 @@ void UpdateDialog::DownloadProgress(size_t downloaded, size_t total)
...
+    // state so they can retry or close the dialog.
+    {
+        wxMessageDialog dlg(this,
+            _("Failed to install the update."),
+            _("Software Update"),
+            wxOK | wxOK_DEFAULT | wxICON_EXCLAMATION);
+        if ( !errorText.empty() )
+            dlg.SetExtendedMessage(wxString(errorText.c_str()));
+        dlg.ShowModal();
+    }
```
vslavik, line 1073: Why didn't you use `StateUpdateError`?

src/msix.cpp
vslavik: Also, is it necessary to do all this on a worker thread — besides the waiting on the async future — or it is just something Claude/Codex slopped out? IOW, which of the operations are time consuming there?

vslavik:
Thanks!

Was this AI-assisted or AI-made? Do you, the human, vouch for the code? Did you test it on pre-Win10?

Requiring explicit installer type is probably the right call for compat reasons... How does this interact with the callbacks set via `win_sparkle_set_can_shutdown_callback` and `win_sparkle_set_shutdown_request_callback`? Are they still called? I assume the 1st one is, the 2nd isn’t? 

Finally, would you be able to provide an example app that could be used for testing?

.github/workflows/build.yml
```diff
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
   build:
     name: Build DLL
-    runs-on: windows-latest
+    runs-on: windows-2022
```
vslavik, line 8: Sorry, how is this related to the PR?

APshenkin: I commented this in commit

windows-latest uses VS 18, and it's not added in .vcxproj (so CI fails to build project then)

I can or add VS18 in .vcxproj or pin to specific version here in ci

vslavik: Sorry, I don’t understand *any* part of this:

> windows-latest uses VS 18, and it's not added in .vcxproj

Why should that matter? What is “not added”? Doesn’t that mean it is something unrelated to this PR that should be separate?

APshenkin:
> Was this AI-assisted or AI-made?

It was AI made (I'm not specialising on Windows stack a lot).

> Do you, the human, vouch for the code?

I read it to see something unusual from my perspective of view, but of course, I can't be 100% sure that it did it in optimal way.

> Did you test it on pre-Win10?

Not yet, my app requires at least Win10, so need to build something simpler to test it.

>  How does this interact with the callbacks set via win_sparkle_set_can_shutdown_callback and win_sparkle_set_shutdown_request_callback?

First one is called, for second one I need to check, but I think it's never called, cause I set `DeploymentOptions::ForceApplicationShutdown`, that terminates application before `RequestShutdown()` is called. But I'll check once again.

> Finally, would you be able to provide an example app that could be used for testing?

yes, I would need to bump some versioning so you can test, will be able to do this next week


*UPDATE*: `RequestShutdown()` is not triggered, if your app was already installed as msix and you process update. However, if there is scenario, that app was originally installed not as msix (different installator, different location where it was installed) and then update is actually msix, then `RequestShutdown()` will be triggered. The reason is that `RegisterApplicationRestart` is only works if you are already in MSIX environment (it takes your packageId, appId and uses them to register restart). So if your app wasn't MSIX and you move to MSIX this callback can be used (for cleanup or something else)

src/ui.cpp
APshenkin: 1. It can't accept error text. DeploymentResult also returns `System.Exception ExtendedErrorCode`, but I don't know how many different codes can be (knowing how Microsoft loves error codes), so enumerate all of them looks not great idea.
2. Possibility of retry. `StateUpdateError` shows only a "Cancel" button and hides the run-installer button. for MSIX A failed install may be recoverable in place as package is already downloaded, e.g. errors related to low disk, conflict process etc
3. Different semantic here, `StateUpdateError` is about failed to find/download update, here it's already downloaded, but failed to install

src/msix.cpp
APshenkin: `deploy::DeploymentResult result = op.get();`

Depends on your app/machine this can take seconds/minutes. And as it's blocking, it will block UI then.

re thread – yes,  will update this

.github/workflows/build.yml
APshenkin: https://github.com/vslavik/winsparkle/actions/runs/28129176634/job/83300765906

CI failed prior the change because, `Winsparkle.vcxproj` contains in configuration
```
<PlatformToolset Condition="'$(VisualStudioVersion)'=='11'">v110_xp</PlatformToolset>
    <PlatformToolset Condition="'$(VisualStudioVersion)'=='12'">v120_xp</PlatformToolset>
    <PlatformToolset Condition="'$(VisualStudioVersion)'=='14'">v140_xp</PlatformToolset>
    <PlatformToolset Condition="'$(VisualStudioVersion)'=='15'">v141_xp</PlatformToolset>
    <PlatformToolset Condition="'$(VisualStudioVersion)'=='16'">v142</PlatformToolset>
    <PlatformToolset Condition="'$(VisualStudioVersion)'=='17'">v143</PlatformToolset>
```

but `windows-latest` has studioversion==18, so as it's not specified there, it fallbacks to v100 by default and it requires then Visual Studio 2010 build tools that not exist in CI environment

this can be separated. Just it fails CI for this PR, that's why changed to pass CI

src/ui.cpp
APshenkin: fixed

APshenkin:
@vslavik answered all questions, fixed thread and operator comments

Here is app with those changes https://api.bron.org/apps/desktop/platforms/win-universal/downloads/1.0.289

If you launch it and wait 5 seconds, popup for update will appear.

I'll try to do somethings for windows 7, but need some free time to setup env for it

vslavik:
> It was AI made (I'm not specialising on Windows stack a lot).

Do you see how that’s problematic? You’re asking for human work on something you don’t understand how or if it works. I agree this would be useful to have, so am willing to put the work in, but the PR is a draft at best — I can’t trust *anything* in it, because I can’t trust that *you* understand it.

I have a new question: how does this interact with UAC/permissions? 

 > I'll try to do somethings for windows 7, but need some free time to setup env for it

Don’t worry, I can do it, have a Win 7 VM readily available.

> Here is app with those changes https://api.bron.org/apps/desktop/platforms/win-universal/downloads/1.0.289

Sorry, what I meant was a test that I could include, i.e. in source code form.

> UPDATE: RequestShutdown() is not triggered, if your app was already installed as msix and you process update. However, if there is scenario, that app was originally installed not as msix (different installator, different location where it was installed) and then update is actually msix, then RequestShutdown() will be triggered. The reason is that RegisterApplicationRestart is only works if you are already in MSIX environment (it takes your packageId, appId and uses them to register restart). So if your app wasn't MSIX and you move to MSIX this callback can be used (for cleanup or something else)

I’m sorry, I don’t understand *anything* in that paragraph. If this PR’s result is that apps can’t reliably be updated using it, then that’s obviously a problem.

eira-fransham:
Wow, I'm currently evaluating WinSparkle for a greenfield application at work, and support for .msix packages looks like it might be the deciding factor. Imagine my surprise when I see that this PR was only posted a week ago.

I wouldn't call myself an expert on WinSparkle or MSIX, but I've been digging into the documentation for both over the past couple of days and I might be able to provide some insight. Sorry for the wall of text, hopefully at least some of it is useful. I'll say upfront that I don't use AI at all, so even if it's word vomit it's at least artisinal word vomit.

> I’m sorry, I don’t understand anything in that paragraph. If this PR’s result is that apps can’t reliably be updated using it, then that’s obviously a problem.

So, my understanding is that updating from a non-MSIX to an MSIX application is not officially supported by Microsoft. MSIX-distributed applications get installed to a different location, get treated differently by the kernel, and have different metadata. I believe that `PackageManager` and other APIs like `ApplicationModel` require that metadata to function correctly, and those APIs cannot be used for non-MSIX applications. As far as I know, it isn't possible to update an MSIX application using the update method that exists in WinSparkle right now because the directory containing the binary is locked down. I'm not sure of the details, and there might be a workaround, but since `PackageManager` is officially-supported I think it would be useful for WinSparkle to provide it as an option. Microsoft expects developers to handle checking for updates etc themselves, they only handle installation.

In my opinion, if an application _really_ wants to convert from being distributed as a standard MSI to MSIX (which is probably not a good idea most of the time anyway), they should treat them like two separate applications. Either direct their users to manually install the MSIX application or have a final update which contains some kind of migration tool to uninstall the old application and install the new one. Then the new MSIX program can start fresh, probably with its own appcast. Non-MSIX applications can keep using the method that currently exists on `master`. There might be a better way, but I don't think it should be WinSparkle's responsibility to make that migration transparent, considering how different MSIX is from the old way of installing applications.

Regarding Windows 7 support, MSIX is technically supported on versions of Windows before 10, but it's not enabled by default. You need to install a separate package, which is officially distributed by Microsoft but it won't be getting official support any more since Windows 7 is EoL. My personal opinion is that it's completely fine to just disable this codepath on Windows 7/8/8.1. Both the existing codepath and the codepath in this PR need to coexist anyway to support non-MSIX applications, and I think the new codepath should be disabled on Windows 7 even for MSIX applications, since the latest version of the MSIX format isn't guaranteed to work on pre-10 Windows.

Finally, I wanted to say that I'd be happy to help out with this PR. I'm reasonably familiar with Windows development and my company is paying for my time to work on figuring out a modern packaging/update solution on Windows. We've already set up Sparkle for our macOS package, so it'd be great if we can use WinSparkle on Windows too.

vslavik:
>  I'll say upfront that I don't use AI at all, so even if it's word vomit it's at least artisinal word vomit.

🤣

> In my opinion, if an application really wants to convert from being distributed as a standard MSI to MSIX (which is probably not a good idea most of the time anyway), they should treat them like two separate applications.

Yes, this is a non-issue. But I think I misread @APshenkin reply when I opened this can of worms (sorry!). For such upgrades, there are two main options:
- Informational-only appcast w/o enclosure, requiring manual update; then using a different appcast feed
- Rely on having an older WinSparkle version that doesn’t understand `installerType` — this I think is good argument for having this kind of explicit opt-in rather than automagic detection of the installer kind.

> Regarding Windows 7 support, MSIX is technically supported on versions of Windows before 10, but it's not enabled by default.

Definitely; I have no intention of supporting MSIX on pre-Windows 10. My concern is entirely that the official `WinSparkle.dll` binary runs on Windows 7 (for now), i.e. that the linkage doesn’t accidentally pull something in. From initial testing, there are some issues with the delayload flags (some superfluous resulting in warnings; missing in release builds), but in principle this works. 

> Finally, I wanted to say that I'd be happy to help out with this PR. I'm reasonably familiar with Windows development and my company is paying for my time to work on figuring out a modern packaging/update solution on Windows.

Great, this is much appreciated! I know next to nothing about MSIX, so any help would be appreciated. In particular a small sample app that I could use for testing (and include in the repo) would be great. And testing, of course.

eira-fransham:
> this I think is good argument for having this kind of explicit opt-in rather than automagic detection of the installer kind.

100% agree. This should be explicitly enabled by the application developer.

> My concern is entirely that the official `WinSparkle.dll` binary runs on Windows 7 (for now), i.e. that the linkage doesn’t accidentally pull something in

Yeah, hopefully that would be mostly addressed by ensuring that it never gets automagically enabled, but I think it could be reasonable to disable the `PackageManager` path by default on versions of Windows without MSIX support built-in. This would apply to Windows 7, 8, 8.1 and versions of Windows 10 prior to 1709. For example, the app dev could set some `MsixUpdate = true` config setting to enable `PackageManager` on Windows 10, but they'd need to _also_ enable something like `MsixCoreSupport = true` to support Windows versions where MSIX doesn't come preinstalled. [MSIX Core](https://learn.microsoft.com/en-us/windows/msix/msix-core/msixcore) is the library that users need to download to allow installing MSIX applications on older versions of Windows.

> In particular a small sample app that I could use for testing (and include in the repo) would be great. And testing, of course.

So I'm currently working on the packaging side of things for our own application, so this should be something I can work on in the next week or two as I start working on updating.

APshenkin:
@vslavik Seems my english is not good enough to describe my thoughts couple of times already and I apologise for that. However, I would try to answer on all compliances to make things cleaner.

> Do you see how that’s problematic? You’re asking for human work on something you don’t understand how or if it works. I agree this would be useful to have, so am willing to put the work in, but the PR is a draft at best — I can’t trust anything in it, because I can’t trust that you understand it.

When I said `It was AI made (I'm not specialising on Windows stack a lot).` I didn't meant that I didn't made any research or not verified code logic, that was produced. I just stated that it was AI made. I know consequences of using AI as well. I understand how the code that was produced works. 

There are few aspects only where I'm not confident enough:
1. Lazy loading of dependencies – the reason is that I just not specialist in Windows programming ecosystem (what I meant in initial reply)
2. I didn't touched wxWidgets before at all, that's why just tried to make things in a same way how they were
3. My main stack is not c++ (even do I had to have do some changes in such projects in the past)

But where I'm confident:
1. MSIX APIs usage – I did reviewed everything what was used
2. MSIX apps bundling and distribution, cause I implemented it by my own in our company year ago
3. COM and winRT APIs – I had experience of them cause working with them a lot in golang based project.

So can I certify what is happening in this PR? – yes, I can. 
Can some elements been improved?  – Probably yes, especially related to places, where I'm not confident.

> I’m sorry, I don’t understand anything in that paragraph. If this PR’s result is that apps can’t reliably be updated using it, then that’s obviously a problem.
First, this PR result to stable reliable update process.

Now, let me try to explain it one more time, so that you can understand this.

Imagine you have app, that was installed on the machine and it **WASN'T** installed via MSIX (msi, NSIS etc). Now, you put MSIX file as update file for it and use `sparkle:installerType="msix"` in your appcast. In this case `RequestShutdown()` **will be** triggered

Now you have the app, that was already installed from MSIX. And you try to update it with MSIX same as in previous example. In that case `RequestShutdown()` **will not be** triggered. The reason for this is because MSIX itself handles restart process then once update is done via  `Management::Deployment`


> In my opinion, if an application really wants to convert from being distributed as a standard MSI to MSIX (which is probably not a good idea most of the time anyway), they should treat them like two separate applications.


> For such upgrades, there are two main options:

> Informational-only appcast w/o enclosure, requiring manual update; then using a different appcast feed
> Rely on having an older WinSparkle version that doesn’t understand installerType — this I think is good argument for having this kind of explicit opt-in rather than automagic detection of the installer kind.

Re usage of MSIX with winsparkle. You should understand (and it's shown in demo videos), that it's possible to use MSIX even now. Just instead of silent install, you will have to click AppInstaller interface to update the app. This is what we did with our app before. With this PR we just wanted to improve user experience to reduce amount of clicks that user should do to update their app.

About updating from `x (take your installer)` to `MSIX` – this is fully on app side. But, I can imagine cases when someone used some installer and `winsparkle` and decided to switch to `msix` (cause msix provides some specific features that are not available without it). And in this case it will be possible
Important to mention that MSIX handles app files and registry differently, rather than usual installations. You can read more here https://learn.microsoft.com/en-us/windows/msix/desktop/desktop-to-uwp-behind-the-scenes#file-system However for backward compatibility you can (as app developer) set MSIX app that way, that it will able to read app state if it was installed before in different way. I know couple of apps as examples who did it that way (1password, slack installer with msix etc).

> Sorry, what I meant was a test that I could include, i.e. in source code form.

Can you then describe in more details, what exactly you want. There is no unit tests in this project, so no references for me what you want.

I can provide you cli commands + AppManifest file,  that allow you to bundle any win32 app in msix, so you can run it then with any app that you have.

vslavik:
> @vslavik Seems my english is not good enough

(Feel free to use DeepL or AI translation/explanation; I have no problem with AI as accessibility tool.)

>  I know consequences of using AI as well. I understand how the code that was produced works.
> MSIX APIs usage – I did reviewed everything what was used

My concern wasn’t whether it generally looks OK (it does), but rather smaller bugs/mistakes. There was enough AI smell that triggered this worry: e.g. the batshit insane way to compare two strings that I think should be noticeable regardless of the language; the duplicate ways of linking libraries (project settings *and* `#pragma`), or the lack of build testing (MSVS build emits warnings about superfluous delayload directives in Debug and fails to delayload in Release).

It’s all right, I appreciate the PR regardless, but it means I absolutely have to give this a lot of human attention.

> Imagine you have app, that was installed on the machine and it WASN'T installed via MSIX (msi, NSIS etc). Now, you put MSIX file as update file for it and use sparkle:installerType="msix" in your appcast. In this case RequestShutdown() will be triggered
> 
> Now you have the app, that was already installed from MSIX. And you try to update it with MSIX same as in previous example. In that case RequestShutdown() will not be triggered. The reason for this is because MSIX itself handles restart process then once update is done via  Management::Deployment

I don’t understand what you’re trying to say here *at all*, sorry. I originally asked whether you designed it so that the shutdown callback is called (requiring the user to make it noop) or not called (requiring docs alert about that). I did *not* expect the answer to be “it is runtime dependent”. 

Please explain why, *specifically*, is the request shutdown somehow called in the unsupported change-installer scenario. This makes no sense to me. Are you saying that failure to install the update results in this? Or what? There’s no specific code for this. What am I missing?

(My opinion on this is that the shutdown callback shouldn’t be called when managed by WinSparkle/MSIX.)

> Can you then describe in more details, what exactly you want. There is no unit tests in this project, so no references for me what you want.

You can’t very well have unit tests for system-altering UAC-using UI... See the `examples` directory for a few simple examples. The point is having something simple that I can run and check updating of.

vslavik:
I’m increasingly more confused by how this is supposed to work.

My understanding was that the system MSIX installer will restart the application automatically via Restart Manager. There’s code to register for it.

But then there’s `UpdateDialog::StateInstallFinished()` which I can’t make sense of:

- How comes it ever gets called except for errors? Isn’t the Windows runtime supposed to restart before that point is reached? Your video certainly looks that way.
- But then why is there code for explicit shutdown?
- Then there’s error message on failure — fair enough, I understand *that*
- But then follows a section for handling "Ready to install.” state — why is it there? What does it do? When is it ever called?

Can you please walk me through how this works in detail? TIA!

